PR #148, opened 2016-10-20
more realistic default binary

### LINKED ISSUE #63: More realistic default binary
The current default binary parameters are unphysical. I think it would be nice to have a much more reasonable default system. Suggestion for parameters below:

sma@binary = 11.0 Rsol (10 would result in 0.75 Msun which is still at least mildly reasonable)
teff@secondary = 6000
teff@primary = 6000

kecnry:
closes #63


kecnry:
closing pull-request for now.  I'll open a new one once we get the new system up and running with nosetest tolerances

PR #174, opened 2016-11-27
The docs link was broken

lethargilistic:
Super minor, but this confused me for a minute, so I figured it was worth fixing.


kecnry:
Thanks for the catch!  I'll merge into master just in case anyone tries to follow the broken link before the upcoming 2.0-release (which will then replace the readme).

Issue #142, opened 2016-10-18
bolometric limb-darkening not passed correctly in the non-convex case


kecnry:
convex case done in 6770f7d8066575ef625ba11dc76215e552277542

added NotImplementedError (see see d904be1835c7d2ef42bf2fcee6e933a05057d50b) for reflection in non-convex cases until this is addressed and well-tested.  We don't currently have any testing cases for reflection in overcontacts (as far as I know)

kecnry:
this will likely need to be handled separately in the development and redistribution branches

kecnry:
see b2b1ca2195474b0a58f3cf74b7d8f2e5f8d28cca and b692eff333451d967d01595829814f08516e59ed

the comparison to legacy is currently failing.  I'm not yet sure if this is a legacy wrapper, pblum, or reflection issue.

PR #181, opened 2016-12-02
rename irradiation effects to match redistribution and handle general case of reflection

### LINKED ISSUE #160: change irradiation parameter names to match those in redistribution branch
before 2.0 we need to settle on final naming conventions.  Currently it seems like the redistribution branch will result in some naming changes, and these need to be adopted into development before the 2.0 branch.

Currently these include:

* reflection_method -> irrad_method
* frac_refl_bol -> irrad_frac_refl_bol

kecnry:
this closes #160 and #142.


kecnry:
I'm going to go ahead and merge this.  The OSX builds are failing but I suspect that is only due to recent reported issues with travis itself.  All of the linux buils (and nosetests) are passing.

Issue #42, opened 2016-08-23
python3 support


kecnry:

kecnry:
created a new branch to prepare and test python3 support https://github.com/phoebe-project/phoebe2/tree/python3_support

Currently building PHOEBE fails.

We're going to need to update (manually) all C-code wrappers.  See here: http://python3porting.com/cextensions.html.  @horvatm - can you do this for libphoebe at some point and I'll try to do it for burlishstoer?

giammarc:
Hi Kyle,

I can help you do that for burlishstoer.  I'm assuming since we went 
back to burlishstoer you'll want v-gamma support at some point?  If that 
is the case I can try to do both at the same time.

I'm assuming this is not highest priority?

-Joe

kecnry:
@giammarc - no not a huge priority, it would just be nice to get everything building on python3 so we can include python3 support in tests.  As for vgamma - that definitely isn't urgent - I'm still hoping to eventually ditch burlishtoer altogether, but we need it for testing against other methods for now.  We should be able to do that testing without vgamma, and if we decide to keep burlishtoer we can deal with fixing the bugs and including vgamma.

kecnry:
current status: we're running into issues with how python3 handles relative imports differently than python2.  I'll look into the best way to handle inter-package imports in a way that should work for both python2 and 3 and then we can continue to see how the nosetests manage.

horvatm:
Python 3 additionally says that this exception cases have invalid syntax:

constraints/expression.py:        except ValueError, msg:
frontend/io.py:                except ValueError, msg:
parameters/parameters.py:        except ValueError, msg:

horvatm:
After commit fc0503b there are still issues with python 3: 

  File "/home/horvat/.local/lib/python3.5/site-packages/phoebe/parameters/constraint.py", line 529, in freq
    rhs = 2 * np.pi / period
TypeError: unsupported operand type(s) for /: 'float' and 'FloatParameter'

It seem that the logic behind algebraic operations between parameters needs fixing.

kecnry:
The above linked commit fixes many issues, including the constraint issues with division (python3 uses \_\_truediv\_\_ and \_\_rtruediv\_\_ instead of \_\_div\_\_ and \_\_rdiv\_\_).  As noted in the commit message, the marshaled passband files are incompatible with Python 3... I will work on trying to find a solution, but if there all else fails, we may need to host two versions of every pasband file: one built with python2 and one with python3.

[...]

PR #323, opened 2019-03-21
Python3 support

kecnry:
support for python 3.6+ and python 2.7+

mpi nosetests are currently disabled for python3 (at least for now) as they cause a timeout issue on travis, but do pass locally.


kecnry:
closes #42

PR #330, opened 2019-05-01
comparison operators hotfix (2.1.7 release)

kecnry:
* Fixes a bug where comparisons between Parameters/ParameterSets and values were returning nonsensical values.
* Comparing ParameterSets with any object will now return a NotImplementedError
* Comparing Parameters will compare against the value or quantity, with default units when applicable.
* Comparing equivalence between two Parameter objects will compare the uniqueids of the Parameters, NOT the values.

### DIFF
--- a/phoebe/parameters/parameters.py
+++ b/phoebe/parameters/parameters.py
@@ -361,6 +361,25 @@ class ParameterSet(object):
 
         return "ParameterSet: {} parameters\n".format(len(self._params))+param_info
 
+    def __lt__(self, other):
+        raise NotImplementedError("comparison operators with ParameterSets are not supported")
+
+    def __le__(self, other):
+        raise NotImplementedError("comparison operators with ParameterSets are not supported")
+
+    def __gt__(self, other):
+        raise NotImplementedError("comparison operators with ParameterSets are not supported")
+
+    def __ge__(self, other):
+        raise NotImplementedError("comparison operators with ParameterSets are not supported")
+
+    def __eq__(self, other):
+        raise NotImplementedError("comparison operators with ParameterSets are not supported")
+
+    def __ne__(self, other):
+        raise NotImplementedError("comparison operators with ParameterSets are not supported")
+
+
     @property
     def meta(self):
         """Dictionary of all meta-tags.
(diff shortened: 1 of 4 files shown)


phoebe/parameters/parameters.py
fazledyn-or, line 1: ## Details
While triaging your project, our bug fixing tool generated the following message(s)-
> In file: parameters.py, class: ParameterSet, there is a special method __gt__ that raises a NotImplementedError. If a special method supporting a [binary operation is not implemented it should return ](https://docs.python.org/3/library/constants.html#NotImplemented)NotImplemented. On the other hand, NotImplementedError should be raised from abstract methods inside user defined base classes to indicate that [derived classes should override those methods. ](https://docs.python.org/3/library/exceptions.html#NotImplementedError)iCR suggested that the special method __gt__ should return NotImplemented instead of raising an exception. An example of how NotImplemented helps the [interpreter support a binary operation is here.](https://docs.python.org/3/library/numbers.html#implementing-the-arithmetic-operations)

Since comparison methods aren't to be supported by the `ParameterSet` class, it's recommended that you use `TypeError` instead of `NotImplementedError`. [Here's](https://docs.python.org/3/library/exceptions.html#TypeError) the reference.

However, if a simple comparison such as below is to be made, both `__eq__` and `__ne__` method is going to raise exception than simply giving the output. In those cases, using `return NotImplemented` will be a better solution.

```py
from phoebe.parameters.parameters import ParameterSet

a = ParameterSet()
b = ParameterSet()
c = a

print(f"a == b, a != b  ::  {a == b}, {a != b}")
print(f"a == c, a != c  ::  {a == c}, {a != c}")
```

## Proposal 
If you're interested, I can create a PR containing these changes. Please let me know.


## Sponsorship and Support
This work is done by the security researchers from OpenRefactory and is supported by the [Open Source Security Foundation (OpenSSF)](https://openssf.org/): [Project Alpha-Omega](https://alpha-omega.dev/). Alpha-Omega is a project partnering with open source software project maintainers to systematically find new, as-yet-undiscovered vulnerabilities in open source code - and get them fixed – to improve global software supply chain security.

The bug is found by running the Intelligent Code Repair (iCR) tool by OpenRefactory and then manually triaging the results.

Issue #542, opened 2021-10-17
Installation (still) fails when m2r is missing


lgov:
When I install the phoebe package - from pip or from current master - in a virtual environment with only numpy installed, the installation fails with this error:

```
Collecting phoebe
  Downloading phoebe-2.3.54.tar.gz (18.6 MB)
     |################################| 18.6 MB 11.2 MB/s            
  Preparing metadata (setup.py) ... error
  ERROR: Command errored out with exit status 1:
   command: /home/lgo/dev/kuleuven/thesis-wip/venv/bin/python3 -c 'import io, os, sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-67xecusn/phoebe_033db77c91e0416ab52b490ec9bb6102/setup.py'"'"'; __file__='"'"'/tmp/pip-install-67xecusn/phoebe_033db77c91e0416ab52b490ec9bb6102/setup.py'"'"';f = getattr(tokenize, '"'"'open'"'"', open)(__file__) if os.path.exists(__file__) else io.StringIO('"'"'from setuptools import setup; setup()'"'"');code = f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-pip-egg-info-o1zycm11
       cwd: /tmp/pip-install-67xecusn/phoebe_033db77c91e0416ab52b490ec9bb6102/
  Complete output (14 lines):
  Traceback (most recent call last):
    File "/tmp/pip-install-67xecusn/phoebe_033db77c91e0416ab52b490ec9bb6102/setup.py", line 340, in <module>
      import m2r
  ModuleNotFoundError: No module named 'm2r'
  
  During handling of the above exception, another exception occurred:
  
  Traceback (most recent call last):
    File "<string>", line 1, in <module>
    File "/tmp/pip-install-67xecusn/phoebe_033db77c91e0416ab52b490ec9bb6102/setup.py", line 343, in <module>
      long_description = fh.read()
    File "/home/lgo/dev/kuleuven/thesis-wip/venv/lib/python3.6/encodings/ascii.py", line 26, in decode
      return codecs.ascii_decode(input, self.errors)[0]
  UnicodeDecodeError: 'ascii' codec can't decode byte 0xc5 in position 2297: ordinal not in range(128)
  ----------------------------------------
```

Information about my setup:
- Ubuntu 18.04.6 LTS
[...]

PR #543, opened 2021-10-17
* setup.py: read README.md file as UTF-8 when m2r is missing.

lgov:
Tested on my Ubuntu system where it solves issue #542 and on Mac where installation still works without m2r.


kecnry:
Thanks @lgov - this looks perfectly reasonable to me... I'll get it merged and prepare a release sometime this week.  Thanks for the PR!

PR #591, opened 2022-05-09
gaussian_process backwards compatibility

kecnry:
This PR tries to consolidate several frontend methods related to gaussian processes while keeping temporary support for syntax for the removed functionality in the upcoming release.

* keep the old `b.add/get/rename_gaussian_process` as opposed to per-kind methods
* `b.add_gaussian_process` accepts `kind` either with or without the `gp_` prefix (and defaults to 'sklearn')
* other methods pass `kind='gp_*'` to the relevant feature-level method
* b.add_feature('gaussian_process') now adds a celerite2 feature.  This is noted as deprecated - so we can remove it in a future release, but might be nice to have for now so the API docs and existing scripts don't suddenly completely break.

@gecheline - do you think these fallbacks (to default to 'sklearn' now, but to have the deprecated old name fallback to 'celerite2' make sense)?  If you don't see any other issues and CI passes, feel free to squash & merge (basic tests seem to do what I'd expect on my machine)!

### DIFF
--- a/phoebe/dependencies/ligeor/eclipse/eb_params.py
+++ b/phoebe/dependencies/ligeor/eclipse/eb_params.py
@@ -1,5 +1,6 @@
 import numpy as np
 from numpy import sin, cos, pi
+import warnings
 from scipy.optimize import newton, minimize
 
 class EbParams(object):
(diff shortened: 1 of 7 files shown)


gecheline:
the add/get_gaussian_process works like a charm. Only thing left to consider is how to handle exclude_phases for GPs. My latest commit has the basic per-dataset implementation.

PR #593, opened 2022-05-12
move gp_exclude_phases parameters to compute context

kecnry:
`gp_exclude_phases` and `gp_exclude_phases_enabled` as per-dataset (via `copy_for`) in the compute context.  `visible_if` statements make these only visible for datasets in which there is an _enabled_ GP feature.

Example use:

```
import phoebe
b = phoebe.default_binary()
b.add_dataset('lc', compute_phases=phoebe.linspace(0,1,101))
```

The `gp_exclude_phases_enabled` and `gp_exclude_phases` parameters are hidden since no GP features have been added

```
print(b.filter(qualifier='gp_*', context='compute'))
b.add_feature('gp_sklearn', dataset='lc01')
print(b['feature'])
```

Now the parameters are visible, in the compute context and per-dataset.  Being in the compute context puts them with the `enabled` parameters and also allows creating different compute options with different options for how to handle the masking.

```
print(b.filter(qualifier='gp_*', context='compute'))
b.filter(qualifier='gp_*', context='compute').twigs
print(b.filter(qualifier='enabled', context='compute'))
```

Adding another feature attached to the same dataset will not make any difference (it will listen to the same parameters as above as they are per-dataset, not per-feature)

```
b.add_feature('gp_sklearn', dataset='lc01')
print(b.filter(qualifier='gp_*', context='compute'))
```

Turning off the enabled parameter hides the list parameter

```
b['gp_exclude_phases_enabled'] = False
print(b.filter(qualifier='gp_*', context='compute'))
b['gp_exclude_phases_enabled'] = True
print(b.filter(qualifier='gp_*', context='compute'))
```

Disabling the features also hides the parameters

```
print(b.filter(qualifier='enabled', feature='gp*', context='compute'))
b.set_value(qualifier='enabled', feature='gp_sklearn01', value=False)
print(b.filter(qualifier='gp_*', context='compute'))
b.set_value(qualifier='enabled', feature='gp_sklearn02', value=False)
print(b.filter(qualifier='gp_*', context='compute'))
b.set_value_all(qualifier='enabled', feature='gp_sklearn02', value=True)
print(b.filter(qualifier='gp_*', context='compute'))
```

Adding another dataset does NOT show more parameters, as the new dataset doesn't have any GPs (yet)

```
b.add_dataset('lc')
print(b.filter(qualifier='gp_*', context='compute'))
```

Once we attach a GP to the new dataset, the new parameters appear in the compute context, tagged with the dataset

```
b.add_feature('gp_sklearn', dataset='lc02')
print(b.filter(qualifier='gp_*', context='compute'))
b.filter(qualifier='gp_*', context='compute').twigs
```

Note: I think I updated the calls that make use of these parameters, but that should be checked against existing scripts.  Existing notebooks will probably need to be updated to add the `gp_` prefix, but as long as they don't use the context (and only refer to qualifier and dataset), they should otherwise work as-is.

### DIFF
--- a/phoebe/frontend/bundle.py
+++ b/phoebe/frontend/bundle.py
@@ -5621,6 +5621,7 @@ class Bundle(ParameterSet):
             self._check_label(kwargs['feature'], allow_overwrite=False)
 
         self._attach_params(params, **metawargs)
+
         # attach params called _check_copy_for, but only on it's own parameterset
         self._check_copy_for()
 
@@ -12045,9 +12046,9 @@ class Bundle(ParameterSet):
                         xqualifier = {'lp': 'wavelength'}.get(ds_ps.kind, 'times')
                         yqualifier = {'lp': 'flux_densities', 'rv': 'rvs', 'lc': 'fluxes'}.get(ds_ps.kind)
                         yerrqualifier = {'lp': 'wavelength'}.get(ds_ps.kind, 'sigmas')
-                        
-                        _exclude_phases_enabled = ds_ps.get_value('exclude_phases_enabled')
-                        
+
+                        _exclude_phases_enabled = computeparams.get_value(qualifier='gp_exclude_phases_enabled', dataset=ds, **_skip_filter_checks)
+
                         if ds_ps.kind in ['lc']:
                             ds_comps = [None]
                         else:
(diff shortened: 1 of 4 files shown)


gecheline:
Didn't get to test it, but I read through the example here and I have no complaints. If that's all you need from me, feel free to merge and close.

kecnry:
Tested against the workshop tutorial and seems to run and return qualitatively equivalent results.

PR #598, opened 2022-05-22
optimizer: differential corrections

kecnry:
This PR implements a simple, single-iteration, differential corrections optimizer solver with support for:
* arbitrary number of LC and RV datasets (requiring per-datapoint sigmas, but no support for other manual weighting)
* compute_times/phases via linear interpolation

The diff is a little messy because of stripping whitespace that I forgot to omit - but we'll want that eventually anyways.  In the "Files Changed" tab, click the setting icon, and then "hide whitespace" to ignore these changes.

### DIFF
--- a/phoebe/parameters/parameters.py
+++ b/phoebe/parameters/parameters.py
@@ -288,6 +288,7 @@ _forbidden_labels += ['primary_width', 'secondary_width',
                       'logwt', 'logl', 'logvol', 'logz', 'logzerr', 'information', 'bound', 'bounds',
                       'bound_iter', 'samples_bound', 'scale',
                       'message', 'nfev', 'niter', 'success', 'initial_values',
+                      'singular_values',
                       'initial_lnlikelihood', 'fitted_lnlikelihood']
 
 
--- a/phoebe/parameters/parameters.py
+++ b/phoebe/parameters/parameters.py
@@ -288,7 +288,7 @@ _forbidden_labels += ['primary_width', 'secondary_width',
                       'logwt', 'logl', 'logvol', 'logz', 'logzerr', 'information', 'bound', 'bounds',
                       'bound_iter', 'samples_bound', 'scale',
                       'message', 'nfev', 'niter', 'success', 'initial_values',
-                      'singular_values',
+                      'singular_values', 'fitted_chi2',
                       'initial_lnlikelihood', 'fitted_lnlikelihood']
 
 
--- a/phoebe/parameters/parameters.py
+++ b/phoebe/parameters/parameters.py
@@ -265,12 +265,13 @@ _forbidden_labels += ['nwalkers', 'niters', 'priors', 'priors_requires', 'init_f
                       'expose_model', 'gtol', 'norm', 'xtol', 'ftol',
                       'priors_combine', 'maxiter', 'maxfev', 'adaptive',
                       'xatol', 'fatol', 'bounds', 'bounds_combine', 'bounds_sigma',
-                      'strategy', 'popsize', 'recombination', 'tol', 'atol', 'polish', 
+                      'strategy', 'popsize', 'recombination', 'tol', 'atol', 'polish',
                       'continue_from', 'continue_from_iter', 'init_from_combine',
                       'burnin_factor', 'thin_factor', 'nlags_factor', 'progress_every_niters',
                       'nlive', 'maxcall', 'lc_geometry', 'rv_geometry', 'lc_periodogram', 'rv_periodogram', 'ebai',
                       'nelder_mead', 'differential_evolution', 'cg', 'powell', 'emcee', 'dynesty',
-                      'analytical_model', 'interactive']
+                      'analytical_model', 'interactive', 'differential_corrections',
+                      'deriv_method', 'steps', 'nsteps']
 
 # from solution:
 _forbidden_labels += ['primary_width', 'secondary_width',
--- a/phoebe/parameters/parameters.py
+++ b/phoebe/parameters/parameters.py
@@ -3300,7 +3300,7 @@ class ParameterSet(object):
 
             if index is not None:
                 if isinstance(param, FloatArrayParameter):
-                    return param.get_value(unit=unit, t=t, **kwargs)[index]
+                    return param.get_value(unit=unit, t=t, **kwargs)[tuple(index)]
                 else:
                     raise ValueError("indices only supported for FloatArrayParameter")
             else:
--- a/phoebe/parameters/solver/optimizer.py
+++ b/phoebe/parameters/solver/optimizer.py
@@ -447,7 +447,8 @@ def differential_corrections(**kwargs):
         model.
     * `expose_lnprobabilities` (bool, optional, default=False): whether to expose
         the initial and final lnprobabilities in the solution (will result in 1
-        additional forward model call)
+        additional forward model call).  Note that since `differential_corrections`
+        does not support priors, this is effectively the lnlikelihood.
     * `continue_from` (string, optional, default='none'): continue the optimization
         run from an existing solution by starting each parameter at its final
         position in the solution.
--- a/phoebe/parameters/solver/optimizer.py
+++ b/phoebe/parameters/solver/optimizer.py
@@ -423,6 +423,8 @@ def differential_corrections(**kwargs):
     Create a <phoebe.parameters.ParameterSet> for solver options for the
     differential-corrections backend.
 
+    This only acts on LC and RV datasets.
+
     Generally, this will be used as an input to the kind argument in
     <phoebe.frontend.bundle.Bundle.add_solver>.  If attaching through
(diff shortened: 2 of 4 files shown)


phoebe/parameters/parameters.py
```diff
@@ -3298,7 +3300,7 @@ def get_value(self, twig=None, unit=None, default=None, t=None, **kwargs):
 
             if index is not None:
                 if isinstance(param, FloatArrayParameter):
-                    return param.get_value(unit=unit, t=t, **kwargs)[index]
+                    return param.get_value(unit=unit, t=t, **kwargs)[tuple(index)]
```
kecnry, line 3303: this avoids a deprecation warning from numpy

phoebe/solverbackends/solverbackends.py
```diff
@@ -2009,13 +2009,6 @@ def mpi_failed_samples_callback(result):
 
 
 class _ScipyOptimizeBaseBackend(BaseSolverBackend):
-    """
-    See <phoebe.parameters.solver.optimizer.nelder_mead>.
-
-    The run method in this class will almost always be called through the bundle, using
-    * <phoebe.frontend.bundle.Bundle.add_solver>
-    * <phoebe.frontend.bundle.Bundle.run_solver>
-    """
```
kecnry, line 2018: moved this into the individual subclasses instead of in this unexposed base class

phoebe/solverbackends/solverbackends.py
```diff
@@ -2101,7 +2094,7 @@ def run_worker(self, b, solver, compute, **kwargs):
                     v = v.to(fitted_units[index]).value
                 p0[p_ind] = v
             else:
-                logger.warning("ignoring {}={} in initial_values as was not found in fit_parameters".format(twig, value))
+                logger.warning("ignoring {}={} in initial_values as was not found in fit_parameters".format(twig_orig, value))
```
kecnry, line 2097: minor logger fix that only affects cases where fit_parameters include an index (so mainly ld_coeffs).  The logger message will now contain the indices.

phoebe/parameters/solver/optimizer.py
```diff
@@ -417,3 +417,73 @@ def differential_evolution(**kwargs):
...
+    for the <phoebe.parameters.compute.phoebe> backend.
+
+    Arguments
+    ----------
+    * `compute` (string, optional): compute options to use for the forward
+        model.
+    * `expose_lnprobabilities` (bool, optional, default=False): whether to expose
+        the initial and final lnprobabilities in the solution (will result in 1
+        additional forward model call).  Note that since `differential_corrections`
+        does not support priors, this is effectively the lnlikelihood.
```
aprsa, line 453: "since" -> ", as"?

phoebe/parameters/solver/optimizer.py
```diff
@@ -417,3 +417,73 @@ def differential_evolution(**kwargs):
...
+    * `continue_from` (string, optional, default='none'): continue the optimization
+        run from an existing solution by starting each parameter at its final
+        position in the solution.
```
aprsa, line 456: is this considered a preferred way to continue iterating? I can easily achieve the same using something like:
```
for iter in range(1, 10):
    b.run_solver(solver='solver_label', solution='solution_label', overwrite=True)
    # any logic to run sanity checks on the solution
    b.adopt_solution(solution='solution_label')
```

phoebe/solverbackends/solverbackends.py
```diff
@@ -2247,6 +2261,240 @@ def valid_options(self):
...
+        return False
+
+    def run_checks(self, b, solver, compute, **kwargs):
+        solver_ps = b.get_solver(solver=solver, **_skip_filter_checks)
+        if not len(solver_ps.get_value(qualifier='fit_parameters', fit_parameters=kwargs.get('fit_parameters', None), expand=True)):
+            raise ValueError("cannot run differential_corrections without any parameters in fit_parameters")
+
+    def _get_packet_and_solution(self, b, solver, **kwargs):
+        # NOTE: b, solver, compute, backend will be added by get_packet_and_solution
+        solution_params = []
```
aprsa, line 2288: Could `solution_params` with all common parameters across all solvers be moved to the base class (`BaseSolverBackend`)?

phoebe/solverbackends/solverbackends.py
```diff
@@ -2247,6 +2261,240 @@ def valid_options(self):
...
+        # limit to LCs/RVs
+        datasets_enabled = b.filter(dataset=datasets_enabled, kind=['lc', 'rv'], context='dataset', **_skip_filter_checks).datasets
+
+        # need to ensure we access the obs and sigma arrays in the same order
+        obs_params = b.filter(qualifier=['fluxes', 'rvs'], dataset=datasets_enabled, context='dataset', **_skip_filter_checks).to_list()
+        obs_data = np.concatenate([p.get_value() for p in obs_params])
+        obs_sigmas = np.concatenate([b.get_parameter(qualifier=['sigmas'], dataset=p.dataset, component=p.component, context='dataset', **_skip_filter_checks).get_value() for p in obs_params])
+
+        if len(obs_sigmas) != len(obs_data):
+            raise ValueError("sigmas must be provided for all enabled lc and rv datasets")
```
aprsa, line 2429: It's quite incredible to see how much boilerplate code is required up to this point. 8-|

phoebe/solverbackends/solverbackends.py
```diff
@@ -2247,6 +2261,240 @@ def valid_options(self):
...
+            obs_baseline = np.array([])
+            for p in obs_params:
+                if p.component == '_default':
+                    continue
+                resid, interp_model = b_solver.calculate_residuals(model=model, dataset=p.dataset, component=p.component, return_interp_model=True, as_quantity=False)
+                xi = np.append(xi, resid)
+                obs_baseline = np.append(obs_baseline, interp_model)
+            return xi, obs_baseline
+
+        baseline_model = b_solver.run_compute(model='baseline', overwrite=True)
```
aprsa, line 2451: Eventually I think we could run this under the hood so that the baseline model is not necessarily exposed.

phoebe/solverbackends/solverbackends.py
```diff
@@ -2247,6 +2261,240 @@ def valid_options(self):
...
+        xi, obs_baseline = _get_residuals_and_interp_model(b_solver, 'baseline', obs_params)
+        if _use_progressbar:
+            _dc_pbar.update(1)
+
+        for k, (uniqueid, value, step) in enumerate(zip(params_uniqueids, p0, steps)):
+            param = b_solver.get_parameter(uniqueid=uniqueid, **_skip_filter_checks)
+            # analytical derivatives:
+            if param.qualifier == 'pblum':
+                A[:,k] = 1
+                continue
```
aprsa, line 2461: is it worth adding another few simple analytical derivatives here or defer that to a later time?

aprsa (approved):
Looks good to me. I left a few comments but none of them critical so I'll approve this now and we can discuss the comments at some point.

phoebe/parameters/solver/optimizer.py
```diff
@@ -417,3 +417,73 @@ def differential_evolution(**kwargs):
...
+    For example:
+
+    ```py
+    b.add_solver('optimizer.differential_corrections')
+    b.run_solver(kind='differential_corrections')
+    ```
+
+    Parallelization support: differential_corrections does not support parallelization.  If
+    within mpi, parallelization will be handled at the compute-level (per-time)
+    for the <phoebe.parameters.compute.phoebe> backend.
```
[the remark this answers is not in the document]
kecnry: The individual forward-models within DC can still be parallelized.

phoebe/parameters/solver/optimizer.py
kecnry: Once you adopt, there is no easy way to "undo".  Using `continue_from` allows you to complete additional iterations without adopting (and is probably also quite a bit cheaper).  Granted, it's more useful for other optimizers where the information to continue is more involved than just starting positions.

phoebe/solverbackends/solverbackends.py
kecnry: Yes, I have a ticket to try to refactor some of this logic.  We have a base class for scipy optimizers, but it will take more work to split out the base logic that applies to other optimizers as well.

phoebe/solverbackends/solverbackends.py
kecnry: None of the models are exposed.  `b_solver` is a copy which is discarded and not accessible to the user (so we don't need to worry about model names conflicting either).

phoebe/solverbackends/solverbackends.py
kecnry: I'd vote to defer.

kecnry:
Ok, as long as it works, I'll merge for now and we can always make small changes (and refactor) later.

PR #635, opened 2022-09-20
negative mass bugfix

### LINKED ISSUE #631: mass parameter has no lower-limit
The mass parameter (generally constrained, but can be flipped to be an input parameter) does not have a defined lower-limit, which allows negative values to be set, causing other constraints to resolve to nans.

See [#630](https://github.com/phoebe-project/phoebe2/discussions/630) for an example.

MichaelAbdul-Masih:
fixed a bug where the mass could be set to negative values.

Fixes #631

### DIFF
--- a/phoebe/frontend/default_bundles/default_binary.bundle
+++ b/phoebe/frontend/default_bundles/default_binary.bundle
@@ -426,8 +426,8 @@ null
 "description": "Source for bolometric limb darkening coefficients (used only for irradiation; 'auto' to interpolate from the applicable table according to the 'atm' parameter, or the name of a specific atmosphere table)",
 "choices": [
 "auto",
-"phoenix",
-"ck2004"
+"ck2004",
+"phoenix"
 ],
 "value": "auto",
 "visible_if": "ld_mode_bol:lookup",
@@ -463,7 +463,7 @@ null
 "value": 0.9988131358058301,
 "default_unit": "solMass",
 "limits": [
-null,
+0.0,
 null
 ],
 "copy_for": false,
@@ -774,8 +774,8 @@ null
 "description": "Source for bolometric limb darkening coefficients (used only for irradiation; 'auto' to interpolate from the applicable table according to the 'atm' parameter, or the name of a specific atmosphere table)",
 "choices": [
 "auto",
-"phoenix",
-"ck2004"
+"ck2004",
+"phoenix"
 ],
 "value": "auto",
 "visible_if": "ld_mode_bol:lookup",
@@ -811,7 +811,7 @@ null
 "value": 0.9988131358058301,
 "default_unit": "solMass",
 "limits": [
-null,
+0.0,
 null
 ],
 "copy_for": false,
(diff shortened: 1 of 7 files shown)


setup.py
```diff
@@ -405,7 +405,7 @@ def _env_variable_bool(key, default):
     long_description = "\n".join(long_description_s[long_description_s.index("INTRODUCTION"):])
 
 setup (name = 'phoebe',
-       version = '2.4.4',
+       version = '2.4.5',
        description = 'PHOEBE devel version',
```
kecnry, line 409: while we're at it, can you update this?

```suggestion
       description = 'PHOEBE 2.4',
```

phoebe/frontend/default_bundles/default_binary.bundle
```diff
@@ -426,8 +426,8 @@ null
 "description": "Source for bolometric limb darkening coefficients (used only for irradiation; 'auto' to interpolate from the applicable table according to the 'atm' parameter, or the name of a specific atmosphere table)",
 "choices": [
 "auto",
-"phoenix",
-"ck2004"
+"ck2004",
+"phoenix"
```
aprsa, line 430: Should these edits be a part of this particular bugfix?

phoebe/frontend/default_bundles/default_binary.bundle
```diff
@@ -463,7 +463,7 @@ null
 "value": 0.9988131358058301,
 "default_unit": "solMass",
 "limits": [
-null,
+0.0,
```
aprsa, line 466: Should this also be set to 1e-12, to be consistent with Kyle's proposed edit in the constraints?

phoebe/frontend/default_bundles/default_contact_binary.bundle
```diff
@@ -463,7 +463,7 @@ null
 "value": 1.0089067994531355,
 "default_unit": "solMass",
 "limits": [
-null,
+0.0,
```
aprsa, line 466: Same as above: 1e-12?

phoebe/frontend/default_bundles/default_contact_binary.bundle
```diff
@@ -811,7 +811,7 @@ null
...
```
aprsa, line 814: Same as above: 1e-12?

phoebe/frontend/default_bundles/default_star.bundle
```diff
@@ -463,7 +463,7 @@ null
 "value": 1.0,
 "default_unit": "solMass",
 "limits": [
-null,
+0.0,
```
aprsa, line 466: Same as above: 1e-12?

aprsa (changes requested):

phoebe/frontend/default_bundles/default_binary.bundle
kecnry: @MichaelAbdul-Masih - these can safely be excluded from the diff if you want, but also don't do any harm (I think the order when exporting to json is somewhat random and so this often shows up when regenerating the default cached bundles).

kecnry (approved):
Looks good to me, thanks!  I'll merge/release once CI passes (it didn't run automatically, but it turns out the triggers weren't setup for a push to a PR, so I updated that and rebased this branch to force it to rerun)

PR #637, opened 2022-09-22
Line profile bugfix (2.4.7 release)

### LINKED ISSUE #607: Line profile wavelength array must include rest wavelength
Because of the subsidiary value definition (and subsequent evaluation in broadening functions), rest wavelength *must* be included in the array of wavelengths. We have several options:
* resample wavelength array to ascertain the inclusion of the rest wavelength;
* instead of using passed wavelengths to compute the profile, compute the profile on an internal hi-res wavelength array and then interpolate the line to the passed wavelength array;
* instead of passing the wavelength array, we should pass wavelength interval width and dispersion.

I like the second option best. Probably something along the lines of:
```
wavelengths = kwargs.get('wavelengths')
wmin, wmax = wavelengths.min(), wavelengths.max()
sigma_w = np.mean(wavelengths[1:]-wavelengths[:-1])
ndown = (profile_rest - wmin) % sigma_w
nup = (wmax - profile_rest) % sigma_w
internal_wls = np.arange(profile_rest - ndown*sigma_w, profile_rest+nup*sigma_w, sigma_w)
# compute profile in internal_wls
fluxes = np.interpolate(wavelengths, internal_wls, internal_fluxes)
```

aprsa:
Line profile dataset takes a wavelength array from the user; when this array does not contain rest wavelength and the subsidiary variable is small, this leads to inadequate sampling and an incorrect line profile. We now use an internal, equidistant wavelength array to calculate the line profile and then interpolate it onto the original wavelength array.

Fixes #607


kecnry:
Definitely would be useful to test against the case in #668 and include that same case from the example notebook in the CI testing and then try to get this out ASAP since I think the change is pretty uncontroversial.

aprsa:
This should now be a complete PR with a test case, version update and a changelog entry.

kecnry:
Before merge we still need to:
- [ ] purge the unused notebooks in tests (`test_l3_flux.comp.model`, `test_l3_fraction.comp.model`)
- [ ] update version in `__init__.py`
- [ ] make sure at least one CI run passes (many are expected to fail until the CI matrix is updated to account for astropy/numpy incompatibilities and the loss of python 3.6 runners)

Raindogjones (approved):

kecnry (approved):
Looks good to me - I'll merge and release tonight, hopefully 🤞 Thanks!

PR #644, opened 2022-10-18
Removes dependency on distutils from setup.py

Raindogjones:
This migrates everything in `setup.py` from `distutils` to `setuptools` and `packaging`.  `setuptools` is a standard package, but `packaging` will need to be added to the prerequisites.


kecnry:
Is it possible that this doesn't listen to `install_requires` anymore (CI is failing since flask isn't being installed, but it is listed in `install_requires`)?  Where do we need to move the dependencies?

Raindogjones:
According to the setuptools docs, install_requires should still be fine...
https://setuptools.pypa.io/en/latest/userguide/quickstart.html#setup-py

kecnry:
reading into the CI log a bit more, it looks like it is processing the dependencies and installing flask but then still giving the error: `error: The 'flask' distribution was not found and is required by phoebe`.  Searching for this online does show that other people/projects have run into similar errors, so we'll have to try some of those proposed solutions.

I can also force CI to re-run on the development branch to see if this is caused by some change introduced in this PR or if its a change in a recent release of flask that is causing us problems.

kecnry:
Awesome, thanks!

> There is now (at least) one failure which I think is due to an incompatibility between astropy=<5 and numpy=>1.23, not sure what to do about this.

The `asscalar` issue existed previously, so as long as that is the only failure, don't worry about it.  I have that on my list to fix, so I'll try to see if I can fix that before we get this in just so that the remainder of those tests can run.

setup.py
```diff
@@ -5,18 +5,31 @@
...
+#   print("Numpy is needed for running and building of PHOEBE")
+#   sys.exit(1)
+
 try:
   import numpy
+  import setuptools
+  import packaging
 except ImportError:
-  print("Numpy is needed for running and building of PHOEBE")
+  print("Numpy, setuptools and packaging are needed for the running and building of PHOEBE")
```
kecnry, line 19: setuptools and packaging should both come with python, right?  I just wonder if we need to actually determine which the user doesn't have and just list those to avoid confusion?

setup.py
```diff
@@ -5,18 +5,31 @@
 """
 import sys
 
+# try:
+#   import numpy
+# except ImportError:
+#   print("Numpy is needed for running and building of PHOEBE")
+#   sys.exit(1)
+
```
kecnry, line 13:
```suggestion
```

setup.py
```diff
@@ -5,18 +5,31 @@
...
 
-from numpy.distutils.core import setup, Extension
-from numpy.distutils.command.build_ext import build_ext
-from numpy.distutils.command.build_py import build_py
+# from numpy.distutils.core import setup, Extension
+# from numpy.distutils.command.build_ext import build_ext
+# from numpy.distutils.command.build_py import build_py
+#
+# from distutils.version import Version, StrictVersion
+# from distutils.cmd import Command
```
kecnry, line 27:
```suggestion
```

kecnry (commented):
@aprsa - any idea why there are a few failing CI builds complaining about the version of pytest?

@Raindogjones - there is an old comment in `setup.py` saying "numpy distutils does not support passing long_description_content_type..." - is this restriction still true with `setuptools`?  If not, maybe we can remove that block of code?  And if so, maybe we should just update the comment so we don't confuse our future selves?

Raindogjones:
> there is an old comment in `setup.py` saying "numpy distutils does not support passing long_description_content_type..." - is this restriction still true with `setuptools`?

@kecnry looks like it is not. I've removed that block, updated the comment and set the `long_description_content_type` in my last commit

aprsa:
@kecnry: off the top of my head I don't know why pytest version would be an issue for astropy; I've not seen this elsewhere. It'll take some digging in to figure out.

aprsa:
I refactored setup.py to try and find the source of the problem; I think the first issue is this:
```
Searching for gevent
Downloading https://files.pythonhosted.org/packages/98/d2/6fa19239ff1ab072af40ebf339acd91fb97f34617c2ee625b8e34bf42393/gevent-websocket-0.10.1.tar.gz#sha256=7eaef32968290c9121f7c35b973e2cc302ffb076d018c9068d2f5ca8b2d85fb0
Best match: gevent websocket-0.10.1
```
The entire output is available [here](https://github.com/aprsa/phoebe2/actions/runs/6867601243/job/18676314033). I _believe_ that this is what @Raindogjones was talking about?

Raindogjones:
@aprsa Exactly! And this goes for the request to install 'flask' installing 'flask-websocket', I think. Hence, I had to specify a version number in order for it to properly resolve.

aprsa:
I was able to work around this issue by installing gevent and friends within the CI, prior to building phoebe. The tests are now passing on linux, but failing on a macos with a cryptic (to me) message:
```
No eggs found in /var/folders/3s/vfzpb5r51gs6y328rmlgzm7c0000gn/T/easy_install-1a4zhmuf/phoebe-2.4.11/egg-dist-tmp-qaxp5les (setup script problem?)
error: The 'phoebe==2.4.11' distribution was not found and is required by the application
Error: Process completed with exit code 1.
```
The full traceback is [here](https://github.com/aprsa/phoebe2/actions/runs/6872430480/job/18690890759).

Raindogjones:
@aprsa I seem to remember seeing this also.  My impression was that CI was installing phoebe from pip as though it was a requirement for phoebe and then getting surprised when the install from source overwrote the pip version. No idea if I managed to fix it though...

aprsa:
Superseded by #796.

PR #654, opened 2022-11-07
Fix handling of extinction grids

Raindogjones:
A number of small fixes to the load and save functions for passbands, in order to match the changes to the way the grids are now stored.

### DIFF
--- a/phoebe/atmospheres/passbands.py
+++ b/phoebe/atmospheres/passbands.py
@@ -425,8 +425,7 @@ class Passband:
             data.append(fits.table_to_hdu(Table({'mu': ck_mus}, meta={'extname': 'CK_MUS'})))
 
         if 'ck2004:ext' in self.content:
-            ck_ebvs = self._ck2004_extinct_axes[-2]
-            ck_rvs = self._ck2004_extinct_axes[-1]
+            ck_teffs, ck_loggs, ck_abuns, ck_ebvs, ck_rvs = self.ext_axes['ck2004']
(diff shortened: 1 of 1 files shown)


kecnry:
Note that a [number of tests](https://github.com/phoebe-project/phoebe2/actions/runs/4109743388/jobs/7091899891) are failing in the `feature-blending` branch and so don't necessarily need to be fixed as part of this PR, but we should ensure that all other tests do pass and that we aren't introducing new points of failure.  Or ideally we should fix `feature-blending` to pass and rebase any additional PRs on top of that once the CI is passing.

Raindogjones:
I think a lot of these failures are a result of Andrej's reworking of passbands.py  

I seem to remember that I managed to track down at least one of the issues and posted it in the slack, but that was now 90+ days ago and so we can't see the message anymore.  If I remember rightly it was some logic issue to do with how extern_planckint atmospheres were being handled...  

I might be able to spend some time tracking it down again later this week, but it would probably be quicker and easier for @aprsa to look as he'd need to make sure any change I propose fits in with...

aprsa:
Yepyep, I already volunteered to do that. It'll take quite some time but we'll get there. :)

kecnry:
Looks like all other tests are passing now that this is on top of #706 and #712  🎉

aprsa (approved):
Looks good to me. At some point I think it'd be worth consolidating all axes into the same array, which probably entails labeling them so that we know what to pull.

PR #674, opened 2022-12-12
update CI matrix

kecnry:
* remove python 3.6 which is past end-of-life and no longer supported by GitHub actions, this also means increasing the lowest tested version of some dependencies which only worked on python 3.6


kecnry:
Replaced by #688

Issue #668, opened 2022-12-08
line profile bug


chipkob:
I'm reporting a bug in the generation of line profiles. I'm following the simple example tutorial at   http://phoebe-project.org/docs/2.4/tutorials/LP

>>> import phoebe
>>> phoebe.__version__
'2.4.5'
>>> b = phoebe.default_binary()
>>> b.add_dataset('lp', times=[0,1,2], wavelengths=phoebe.linspace(549, 551, 101))
<ParameterSet: 100 parameters | contexts: compute, dataset, figure, constraint>
>>> b.run_compute(irrad_method='none')
<ParameterSet: 5 parameters | qualifiers: comments, flux_densities, wavelengths>
>>> print(b.filter(qualifier='flux_densities', context='model'))
ParameterSet: 3 parameters
R  00.000000@flux_densities@la...: [1. 1. 1. ... 1. 1. 1.] W / (m2 nm)
R  01.000000@flux_densities@la...: [nan nan nan ... nan nan nan] W / (m2 nm)
R  02.000000@flux_densities@la...: [nan nan nan ... nan nan nan] W / (m2 nm)

See that the latter two time entries return nan whereas the example output on the tutorial page shows 1. 1. 1.  
Something seems amiss.  This error also happens in phoebe 2.4.2.    phoebe 2.2.2 appears to function properly in this regard, however.

kecnry:
Thanks for the report.  Although it isn't exactly the same, it is likely that #637 will fix this.  Feel free to try that branch and see if it works for you, or we will just try to double check this case before releasing that.

aprsa:
I can confirm that #637 fixes this.

kecnry:
#637 fixes this and is included in the 2.4.7 release.

chipkob:
Hmm, I still get nans for all line profiles after the first.  My terminal transcript follows.  There must be something else involved...
Chip 


Successfully installed phoebe-2.4.7
(base) bash-4.4$ python
Python 3.9.12 (main, Apr  5 2022, 06:56:58) 
[GCC 7.5.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import phoebe
>>> b = phoebe.default_binary()
>>> b.add_dataset('lp', times=[0,1,2], wavelengths=phoebe.linspace(549, 551, 101))
<ParameterSet: 100 parameters | contexts: compute, figure, constraint, dataset>
[...]

PR #678, opened 2022-12-16
Fixed a wavelength coverage bug and added a new test.

aprsa:
This first test incorrectly set the wavelength span from 0 to 1 instead of 549 to 551, which has been fixed. The new test checks whether internal interpolation works as expected.


kecnry (approved):
Seems reasonable and all the CI runners that used to pass still seem to be passing (I'm tempted to remove the twigs to save a bit of time for the CI, but for one iteration it's probably negligible).  Let's get this in and then we may need to revise them further to address the compatibility issues with older version of numpy in #668 as a follow-up, especially once the CI matrix is fixed.  Thanks!

Note: this will be merged into main but since it only affects the tests will not get its own bugfix release (and will just be included in the next release).

Issue #687, opened 2023-02-01
Astropy fails to build within CI python 3.8


aprsa:
Specific versions:

Python 3.8.*
Scipy 1.8.*
Numpy 1.18.*
Astropy 3.0.*
Matplotlib 3.5.*

Astropy fails to build.

aprsa:
Confirming the same failure on macos-12.

Issue #686, opened 2023-01-31
Mac-os 11 builds fail with astropy 2.x


aprsa:
As part of CI, we test python 3.7 with astropy [2-4], numpy [1.15-1.21], scipy [1.4-1.7]. All tests work on os=ubuntu-latest, but os=macos-11 with astropy=2 fail because pip fails to build astropy. These are explicitly excluded in the CI yml file but we should check to see if anything can be done to still salvage these builds. Ooooor wait for 3.7 to reach end-of-life and not worry about it. ;)

PR #688, opened 2023-02-02
CI rewrite

### LINKED ISSUE #685: Nosetests deprecated, replace with pytest
Nose tests have not been maintained and have been replaced with pytest. We need to convert our test suite to pytest. See:

https://docs.pytest.org/en/7.2.x/how-to/nose.html

aprsa:
This PR establishes a new set of github actions rules for CI testing. A couple of tests are still failing on the phoebe side and need to be investigated. Others are setup issues and tickets have been opened for those as there is no apparent reason why the particular combinations of software versions aren't expected to work (see #686 and #687). Coverage has been expanded and covers python 3.7 through 3.11, along with suitable versions of numpy, scipy, astropy and matplotlib. In the interest of time, however, coverage has been reduced to a manageable size. If we decide to support full coverage, we will need to run tests on our end instead of github, where we are limited to 1000 minutes.

Closes #685

### DIFF
--- a/.github/workflows/ci-nosetests.yml
+++ b/.github/workflows/ci-nosetests.yml
@@ -63,41 +63,41 @@ jobs:
     # #     auto-update-conda: true
     # #     python-version: ${{ matrix.python-version }}
 
-    # - name: Setup python ${{ matrix.python-version }}
-    #   uses: actions/setup-python@v2
-    #   with:
-    #     python-version: ${{ matrix.python-version }}
+    - name: Setup python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
 
-    # - name: MacOS Xcode setup
-    #   if: matrix.os == 'macOS-10.15'
-    #   uses: maxim-lobanov/setup-xcode@v1
-    #   with:
-    #     xcode-version: latest-stable
+    - name: MacOS Xcode setup
+      if: matrix.os == 'macOS-10.15'
+      uses: maxim-lobanov/setup-xcode@v1
+      with:
+        xcode-version: latest-stable
 
-    # - name: MacOS Setup # https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md
-    #   if: matrix.os == 'macOS-10.15'
-    #   run: |
-    #     export CPATH=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include
+    - name: MacOS Setup # https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md
+      if: matrix.os == 'macOS-10.15'
+      run: |
+        export CPATH=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include
 
     # - name: PHOEBE environment variables
     #   run: |
     #     export PHOEBE_ENABLE_PLOTTING='FALSE'
     #     export PHOEBE_UPDATE_PASSBAND_IGNORE_VERSION='TRUE'
 
-    # - name: Update pip
-    #   run: |
-    #     python -m pip install --upgrade pip
+    - name: Update pip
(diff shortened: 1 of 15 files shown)


kecnry:
Thanks - this will be great to finally have modernized and running successfully!  Few thoughts:
* Do we really want to merge with 100 tests while running on GitHub runners?  If not, let's trim it down in a single revertible commit (either squash all of the existing commits into one and we'll do a normal merge or just make sure not to delete your branch so we still have access to it when doing a squash & merge) and create a follow-up issue to re-enable the removed tests.
* Similarly, I would also vote to remove the failing tests in a revertible commit so that when we rebase other PRs we don't need to remember which are expected to fail.
* Do we want to re-enable benchmarks, re-enable in a follow-up effort, or abandon it?  
* I don't see any harm in deleting the commented-out blocks that we don't need anymore, they'll still live in the git history in this PR if we ever need to reference/re-introduce them.

aprsa:
Here are my thoughts:

- my vote would be to keep the current matrix for pull requests, and trim it to something smaller, say 1 per python version per os, for pushes;
- one of the failing tests is phoebe; we need to look into that one; the other two failing tests are mac builds. Someone with a mac should be able to debug that better than I can. We can of course open tickets and then exclude them from the runs, but I think it's worth the effort to just fix them now;
- I haven't touched the benchmarks at all; I'd certainly not have them in the same yaml. If anything, I'd move them to a benchmarks file that's run on-demand rather than automatically;
- I left the commented out bits there to help you see what's been changed and how. I'm happy to see those removed.

kecnry:
> my vote would be to keep the current matrix for pull requests, and trim it to something smaller, say 1 per python version per os, for pushes

We can try that, I worry we might still go over our quotas, but let's find out!

> We can of course open tickets and then exclude them from the runs, but I think it's worth the effort to just fix them now

Ok, so are you suggesting including those fixes in this PR (and waiting to merge)?  Otherwise, I'd really like CI to be green when we do merge so that we can move forward on other PRs without having to remember which are known/expected to fail and which aren't.  I agree that addressing those should be pretty high priority though.

> If anything, I'd move them to a benchmarks file that's run on-demand rather than automatically

I like that idea, and probably only on PR, not push?  Right now we don't actually enforce anything, but its at least nice to have them logged so we can keep an eye on trends and more easily find the blame for a decrease in performance.

> I left the commented out bits there to help you see what's been changed and how. I'm happy to see those removed.

Ok, let's cut them then, they're still in the diff.

aprsa:
>> We can of course open tickets and then exclude them from the runs, but I think it's worth the effort to just fix them now

> Ok, so are you suggesting including those fixes in this PR (and waiting to merge)? Otherwise, I'd really like CI to be green when we do merge so that we can move forward on other PRs without having to remember which are known/expected to fail and which aren't. I agree that addressing those should be pretty high priority though.

We can include it as a part of this PR if it's obvious; I expect mac builds to be an easy fix. However, test_rm is failing on a mac run and that might take effort to track down. Perhaps you can take a glance and see if you recognize those issues? If not, we can exclude those runs and add high priority tickets for them?

>> If anything, I'd move them to a benchmarks file that's run on-demand rather than automatically

> I like that idea, and probably only on PR, not push?

Even PR might be an overkill. There's no pushing to the repository anyway, right? So it kind-of defeats the purpose -- we might simply leave it as-is and it would do what we want anyway?

kecnry:
ok, I agree, let's try to get the mac build fixed here before merge, if possible.  `test_rm` looks to be more involved and I think should be a separate PR even if we know how to fix it now, since it likely will need to be some change to logic within phoebe and I don't want that to get buried in this large PR.

> Even PR might be an overkill. There's no pushing to the repository anyway, right? So it kind-of defeats the purpose -- we might simply leave it as-is and it would do what we want anyway?

Actually that's true - since it only runs for specific branches, those shouldn't exist in forks and so should only trigger on a PR merge into one of those branches.  So let's either enable it for latest-everything or as its own workflow that runs on latest-everything?

kecnry:
Follow-up issues from failing builds that have been excluded here: #686, #687, #689.

kecnry (approved):
Looks good to me to merge once CI is green, thanks!

Issue #638, opened 2022-10-06
How do I handle software installation problems


Yzxastr:
HI,i ran into this bug when I installed it in anaconda by 'pip install numpy phoebe ' or 'pip3 install numpy phoebe ' . I have installed the latest Visual Studio 2022 and Microsoft Visual C++ Redistributable.
My bug is as follows:
![bug1](https://user-images.githubusercontent.com/102363557/194208646-b44be68b-b6cf-47cc-83a5-4f93402641
![bug2](https://user-images.githubusercontent.com/102363557/194208651-ce2be573-0922-4bac-93d1-335201fd402a.png)
5c.png)

Could you give me some suggestions to solve this bug?

Yzxastr:
![bug1](https://user-images.githubusercontent.com/102363557/194208848-e0e44e4b-5d7f-4616-9328-7c9cac0b0f1b.png)

kecnry:
Sorry you're running into issues!  We don't have too much experience with windows systems ourselves, so would definitely consider support for windows somewhat experimental and are trying to revise the installation instructions based on what works for users.

Just to try to get a few more details on what might be different between what you tried and what has worked for other users:
* Did you also install the "coding pack for python" listed on the [windows install page](http://phoebe-project.org/install/latest/windows/auto)?  
* Are you using a conda environment or trying to install without?

Thanks
Kyle

horvatm:
Did installation of numpy via
`  pip install numpy`
succeed?

Cheers, Martin

Yzxastr:
succeed




------------------&nbsp;原始邮件&nbsp;------------------
发件人: "Martin ***@***.***&gt;; 
发送时间: 2022年10月7日(星期五) 晚上10:03
收件人: ***@***.***&gt;; 
抄送: ***@***.***&gt;; ***@***.***&gt;; 
主题: Re: [phoebe-project/phoebe2] How do I handle software installation problems (Issue #638)





 
Did installation of numpy via
   pip install numpy
 succeed?
 
Cheers, Martin

Yzxastr:
I have installed the "coding pack for python", and i used a conda environment to install phoebe by "pip install numpy phoebe".






------------------&nbsp;原始邮件&nbsp;------------------
发件人:                                                                                                                        "phoebe-project/phoebe2"                                                                                    ***@***.***&gt;;
发送时间:&nbsp;2022年10月7日(星期五) 晚上9:13
***@***.***&gt;;
[...]

PR #691, opened 2023-02-07
recompute teffs for spots in async case

kecnry:
the mesh does not need to be recomputed for spots (as changed in 2.4.8 release in #638), but if the rotation is asynchronous, then the surface temperatures *do* need to be recomputed.

### DIFF
--- a/phoebe/backend/universe.py
+++ b/phoebe/backend/universe.py
@@ -1998,7 +1998,9 @@ class Star_roche(Star):
 
     @property
     def needs_recompute_instantaneous(self):
-        return self.needs_remesh
+        # recompute instantaneous for asynchronous spots, even if meshing
+        # doesn't need to be recomputed
+        return self.needs_remesh or (len(self.features) and self.F != 1.0)
 
     @property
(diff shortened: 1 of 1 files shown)


phoebe/backend/universe.py
```diff
@@ -1998,7 +1998,9 @@ def is_convex(self):
 
     @property
     def needs_recompute_instantaneous(self):
-        return self.needs_remesh
+        # recompute instantaneous for asynchronous spots, even if meshing
+        # doesn't need to be recomputed
```
aprsa, line 2002: while this plugs the leak, would it not be better to handle async spots at the teff level rather than recomputing the entire mesh?

aprsa (approved):
Looks good to me. Should we add a small test for mesh recomputation at some point soon?

phoebe/backend/universe.py
kecnry: yes, agreed, I created a follow-up ticket to revisit this entire remeshing/computing infrastructure so we can remain as efficient as possible without breaking any assumptions.

PR #694, opened 2023-02-09
Implementation of gravitational redshift fixed.

### LINKED ISSUE #608: rv_grav units/order of magnitude mismatch
A Sun-Jupiter system with rv_grav enabled returns orders of magnitude higher RVs for the Sun than expected (~600 m/s expected, ~700 km/s returned)

A single Sun-like star has the correct rv_grav offset though.

aprsa:
The radius used in computing the gravitational redshift was missing the semi-major axis. That caused phoebe to incorrectly compute the gravitational redshift correction when rv_grav=True was passed to run_compute(). The computation has been fixed and a unit test has been added. This closes #608.

### DIFF
--- a/phoebe/backend/universe.py
+++ b/phoebe/backend/universe.py
@@ -1734,11 +1734,14 @@ class Star(Body):
 
         rvs = -1*self.mesh.velocities.for_computations[:,2]
 
-
         # Gravitational redshift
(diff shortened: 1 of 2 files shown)


kecnry (approved):
Works for me, thanks!

PR #701, opened 2023-02-15
Fixed two bugs with JKTEBOP backend RVs.

### LINKED ISSUE #699: JKTEBOP backend RV semi-amplitudes not working with mass ratio != 1
Minimum working example included as a jupyter notebook in attachment. [rv_jktebop_example.zip](https://github.com/phoebe-project/phoebe2/files/10732915/rv_jktebop_example.zip)

Hi!

I ran across this issue when I wanted to make a forward model in PHOEBE using parameters from my (external) JKTEBOP fit. I tried using the JKTEBOP (internal) backend to compare, and it did not match very well at all.

I can reproduce this problem by starting with q=M2/M1=1, where the PHOEBE and JKTEBOP backend produces equivalent RV curves. Afterwards, when altering with q=0.5, the PHOEBE model is changed. But the JKTEBOP model produces the exact same result as before.

It looks like the relevant line is:
https://github.com/phoebe-project/phoebe2/blob/c4e2e30d5222a01fc9a8bc2831145093841d09d4/phoebe/backend/backends.py#L2340

As far as I can see, calculation of the RV semi-amplitudes is using the system 'sma@binary' in the JKTEBOP backend, instead of 'sma@primary' and 'sma@secondary'?

jsinkbaek:
1. To calculate RV semi-amplitude (K), system semi-major-axis was used previously instead of component sma (sma vs. center of mass). Before, any change in the mass ratio had no effect on the modelled rvs.

2. Small left-over discrepancy corrected, as RV amplitude K did not take eccentricity into account. Now includes term 1/(sqrt(1-ecc^2)).

Closes #699


kecnry (commented):
Changes to the code look reasonable to me, thanks!  We just need to incorporate tests on our end before approving/merging.

kecnry (approved):
Thanks for the fixes @jsinkbaek and the test coverage @MichaelAbdul-Masih!  I'll merge this once all the tests finish running (assuming they pass) and then it should make it in the next bugfix release (2.4.11).

PR #706, opened 2023-02-16
Bugfixes to get CI to pass for blending.

### LINKED ISSUE #31: intensities are unnecessarily computed twice for RVs
If an RV is attached to both stars in a binary, then the intensities are computed twice.  This should be an easy place to optimize RV computations.

aprsa:
This is a fairly large and heterogeneous commit that gets all CI tests to pass. Worthy of mention are the following changes:

* the default value for parameter `blending_method` changed from 'linear' to 'none';
* the default value for parameter `ld_blending_method` changed from 'nearest' to 'none';
* default bundles have been recomputed to account for the change in parameter defaults;
* the visibility for parameter `ld_blending_method` no longer depends on `blending_method`;
* fallback value was added for ld_extrapolation_method in bundle.py for computes where kind != 'phoebe';
* blending parameters are set to 'none' for atm='blackbody' to avoid later problems;
* string representations for ComputedColumn class added (I needed them for debugging and deemed them useful so I left them there);
* a bunch of small fixes in passbands.py, mostly do to with array indexing.

### DIFF
--- a/phoebe/parameters/compute.py
+++ b/phoebe/parameters/compute.py
@@ -82,9 +82,9 @@ def phoebe(**kwargs):
     * `atm` (string, optional, default='ck2004'): atmosphere table
     * `irrad_method` (string, optional, default='horvat'): which method to use
         to handle irradiation.
-    * `blending_method` (string, optional, default='linear'): Method to use for
+    * `blending_method` (string, optional, default='none'): Method to use for
         blending. (Only applicable if `atm` is not 'blackbody')
-    * `ld_blending_method` (string, optional, default='nearest'): Method to use
+    * `ld_blending_method` (string, optional, default='none'): Method to use
         for extrapolating limb-darkening during blending (for all datasets and
         bolometric for irradiation, if applicable).  (Only applicable if `atm`
         is not 'blackbody' and `blending_method` is not 'none')
@@ -177,8 +177,8 @@ def phoebe(**kwargs):
 
     # PER-COMPONENT
     params += [ChoiceParameter(copy_for = {'kind': ['star'], 'component': '*'}, component='_default', qualifier='atm', value=kwargs.get('atm', 'ck2004'), choices=_atm_choices, description='Atmosphere table')]
-    params += [ChoiceParameter(visible_if='atm:!blackbody', copy_for = {'kind': ['star'], 'component': '*'}, component='_default', qualifier='blending_method', value=kwargs.get('blending_method', 'linear'), choices=['none', 'nearest', 'linear'], description='Method to use for blending')]
-    params += [ChoiceParameter(visible_if='atm:!blackbody,blending_method:!none', copy_for = {'kind': ['star'], 'component': '*'}, component='_default', qualifier='ld_blending_method', value=kwargs.get('ld_blending_method', 'nearest'), choices=['none', 'nearest', 'linear'], description='Method to use for extrapolating limb-darkening during blending (for all datasets and bolometric for irradiation, if applicable)')]
+    params += [ChoiceParameter(visible_if='atm:!blackbody', copy_for = {'kind': ['star'], 'component': '*'}, component='_default', qualifier='blending_method', value=kwargs.get('blending_method', 'none'), choices=['none', 'nearest', 'linear'], description='Method to use for blending')]
+    params += [ChoiceParameter(visible_if='atm:!blackbody', copy_for = {'kind': ['star'], 'component': '*'}, component='_default', qualifier='ld_blending_method', value=kwargs.get('ld_blending_method', 'none'), choices=['none', 'nearest', 'linear'], description='Method to use for extrapolating limb-darkening during blending (for all datasets and bolometric for irradiation, if applicable)')]
 
     # PER-DATASET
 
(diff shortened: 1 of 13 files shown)


tests/nosetests/test_sun_earth/test_sun_earth.py
```diff
@@ -33,6 +33,7 @@ def initiate_sun_earth_system(pb_str, atm='blackbody'):
 
     b.set_value_all('irrad_method', 'none')
     b.set_value_all('blending_method', value='none')
+    # b.set_value_all('ld_blending_method', value='none')
```
kecnry, line 36:
```suggestion
```

tests/nosetests/test_legacy_parser/test_legacy_export_import.py
```diff
@@ -38,18 +32,15 @@ def test_reimport(filename=None):
...
             
             try:
-                assert(val1==val2)
+                assert val1 == val2
             except:
-                assert(all(val1==val2))
-
-
-
+                assert all(val1==val2)
```
kecnry, line 37: not necessary here... but why is there a try/except?  Can we just check for array vs float explicitly?

tests/nosetests/test_ld/test_ld.py
```diff
@@ -204,4 +199,4 @@ def test_binary(plot=False, gen_comp=False):
     logger = phoebe.logger(clevel='INFO')
 
 
-    b = test_binary(plot=False, gen_comp=True)
+    b = test_binary(plot=True, gen_comp=True)
```
kecnry, line 202:
```suggestion
    b = test_binary(plot=plot, gen_comp=True)
```

phoebe/parameters/compute.py
```diff
@@ -177,8 +177,8 @@ def phoebe(**kwargs):
...
-    params += [ChoiceParameter(visible_if='atm:!blackbody,blending_method:!none', copy_for = {'kind': ['star'], 'component': '*'}, component='_default', qualifier='ld_blending_method', value=kwargs.get('ld_blending_method', 'nearest'), choices=['none', 'nearest', 'linear'], description='Method to use for extrapolating limb-darkening during blending (for all datasets and bolometric for irradiation, if applicable)')]
+    params += [ChoiceParameter(visible_if='atm:!blackbody', copy_for = {'kind': ['star'], 'component': '*'}, component='_default', qualifier='blending_method', value=kwargs.get('blending_method', 'none'), choices=['none', 'nearest', 'linear'], description='Method to use for blending')]
+    params += [ChoiceParameter(visible_if='atm:!blackbody', copy_for = {'kind': ['star'], 'component': '*'}, component='_default', qualifier='ld_blending_method', value=kwargs.get('ld_blending_method', 'none'), choices=['none', 'nearest', 'linear'], description='Method to use for extrapolating limb-darkening during blending (for all datasets and bolometric for irradiation, if applicable)')]
```
kecnry, line 181: is 'none' really the default that we want now?  Or just what is needed for the legacy wrapper?

kecnry (commented):
Few small comments before I can do a more thorough review.

I do think we should try to move as many of these stylistic changes that don't affect blending into `release-2.5` (or perhaps a bugfix) to avoid conflicts down the road.

Raindogjones:
I've gone through the changes in passbands.py and it all looks good to me.  We'll still need to fix the load and save functions, but those are in #654

aprsa:
A single CI test, `test_backwards_compatibility`, is failing. The failure is linked to the Johnson:R and TESS:T passbands. I cannot reproduce this on my end no matter what I do. @kecnry, any chance you can try to reproduce? Would CI for any reason not be able to pull these passbands properly from the server? It's failing on the ck_teffs key that is certainly in these files.

phoebe/parameters/compute.py
aprsa: This merits a longer discussion. Should extrapolation and blending be disabled or enabled by default?

kecnry:
The server is still serving passband files for phoebe up to version 2.4, so we will need to either update the server to also support (future) version 2.5 **and/or** add the capability in phoebe to handle the older format on the fly. If you mean they it already should be able to do that, then can you reproduce locally by clearing your version of those passbands and running the tests?

phoebe/parameters/compute.py
Raindogjones: For me, it's clear that blending/extrapolation should be disabled by default.

aprsa:
The upcoming 2.5 version is backwards-compatible with the currently served 2.4 passbands -- they should just load, and they do on my side (after being pulled from the server), but do not on CI. That's why I asked if you could reproduce the failure because I cannot.

kecnry:
Yes, I can reproduce once removing my locally installed versions of those passbands and forcing the test to pull them from the server.

kecnry:
Let's not have the backwards compatibility tests hold up approving/merging this PR - those will require the passband server to be updated to use this version of phoebe and we have confirmed that it runs locally with the full passband file.  That will show all tests to fail though... so maybe we should switch the passband in those tests to one of the shipped passband files and switch it back before release to make sure we have the test coverage of pulling from the server?

PR #707, opened 2023-02-21
Server-side fix for the missing supporting tables

aprsa:
This fix corrects the packing of supporting tables for the upcoming 2.5 release.

### DIFF
--- a/phoebe/atmospheres/passbands.py
+++ b/phoebe/atmospheres/passbands.py
@@ -310,7 +310,7 @@ class Passband:
         data.append(primary_hdu)
 
         # Tables:
-        ptf_table = Table(self.ptf_table)
+        atms = np.unique([content.split(':')[0] for content in self.content])
         data.append(fits.table_to_hdu(Table(self.ptf_table, meta={'extname': 'PTFTABLE'})))
 
         if 'blackbody:Inorm' in self.content:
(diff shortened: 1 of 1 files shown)


phoebe/atmospheres/passbands.py
```diff
@@ -310,7 +310,7 @@ def save(self, archive, overwrite=True, update_timestamp=True, history_entry='')
         data.append(primary_hdu)
 
         # Tables:
-        ptf_table = Table(self.ptf_table)
```
kecnry, line 313: was this just completely unused in this method?

kecnry (commented):
All these changes seem completely backwards-compatible and I can't think of any user-workflow that would be affected, so I'm fine with including this in the next bugfix so the passband-server can be better future-proofed.  We probably will want to test with a local version of the server just to make sure it achieves everything we think it should first though.

phoebe/atmospheres/passbands.py
aprsa: yep, completely unused.

kecnry:
I added instructions to the [passband server](https://github.com/phoebe-project/tables.phoebe-project.org) for how to test locally.  Before merging, one of us will need to test that this works as expected to fix the failing test in #706 by running the server on this branch, and a local version of phoebe (in a separate environment) with both 2.4.x and the branch from #706, with the local versions of `TESS:T` and `Johnson:R` removed, and make sure the files are hosted and parsed correctly in both cases.

Assuming that passes, this can make the next bugfix release, and then we will need to update the version running on the passband server itself to the new bugfix, and then CI should hopefully pass :crossed_fingers:.  I don't think this necessarily needs to hold up #706 though - we can either move forward with that in the meantime and just manually check CI logs to make sure that is the only failing test on other PRs or temporarily changes those tests to use the shipped passbands.

aprsa:
Checks are now passing on the updated server, and everything I tried on my end (with phoebe 2.4) works as expected. We should test on phoebe 2.2 and 2.3, and then I think this is ready to go in.

PR #712, opened 2023-03-02
Un"fixing" passband tables

aprsa:
In an attempt to address missing auxiliary tables in served passbands, the check for non-Imu atmospheres was inadvertantly removed. This commit unfixes that.

### DIFF
--- a/phoebe/frontend/bundle.py
+++ b/phoebe/frontend/bundle.py
@@ -3799,7 +3799,7 @@ class Bundle(ParameterSet):
                                         [pbparam, atmparam],
                                         True)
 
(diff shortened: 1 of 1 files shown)


kecnry:
Changes look simple enough - I'll approve and merge once CI shows all tests except backwards compatibility passing and can then rebase the other PRs on top of this.  Thanks!

aprsa:
Sounds good to me -- this will fix everything except the backwards compatibility test, which needs #707 merged into master.

kecnry (approved):
CI seems to be showing all expected tests passing (just from a random sample of all the CI runs).  Let's get this in so we can rebase the other.  Thanks!

PR #715, opened 2023-03-13
Adding MacOS-12 to CI

aprsa:
This PR adds macos-12 to the CI matrix.

### DIFF
--- a/.github/workflows/on_pr_test_python_31x.yml
+++ b/.github/workflows/on_pr_test_python_31x.yml
@@ -6,6 +6,7 @@ on:
   pull_request:
   push:
     branches:
+      - 'macos12_ci'
       - 'master'
       - 'feature-*'
       - 'bugfix-*'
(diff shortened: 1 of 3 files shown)


.github/workflows/on_pr_test_python_31x.yml
```diff
@@ -6,6 +6,7 @@ on:
   pull_request:
   push:
     branches:
+      - 'macos12_ci'
```
kecnry, line 9: This PR itself should still run CI without needing to list it in the list of push branches, right?

```suggestion
```

.github/workflows/on_pr_test_python_38.yml
```diff
@@ -6,6 +6,7 @@ on:
   pull_request:
   push:
     branches:
+      - 'macos12_ci'
```
kecnry, line 9:
```suggestion
```

.github/workflows/on_pr_test_python_39.yml
```diff
@@ -6,6 +6,7 @@ on:
   pull_request:
   push:
     branches:
+      - 'macos12_ci'
```
kecnry, line 9:
```suggestion
```

kecnry (commented):
This shouldn't currently cause conflicts with the bugfix release, so is probably fine to merge directly into master.  We'll only need to rebase the bugfix branch if we are trying to address the failure there.

aprsa:
Just for the record, I canceled all tests for the last 3 commits because the changes were inconsequential.

PR #729, opened 2023-04-27
Adds option to flip component asini constraint solving for q

Raindogjones:


kecnry:
I'll need to think whether this qualifies for a bugfix release or needs to be in the next feature release.  I think since this is in the constraint function definition, a bundle saved before this change and loaded after should work, but it might be worth checking.

I do think we should do the same for `sma@star` while we're at it - that should essentially be the exact same logic and hopefully won't conflict with each other 🤞

Raindogjones:
Ok, I'll try to add that too.

kecnry:
We'll also need to update the `_validsolvefor` defined directly above the functions as well.  It looks like the API docs don't explicitly mention what parameters are allowed/not supported, so I guess we don't need to worry about that for now which will make including in a bugfix a bit easier.

Thanks!

PR #736, opened 2023-05-18
Bring back boosting!

Raindogjones:
closes phoebe-project/phoebe-development/issues/54


phoebe/frontend/bundle.py
```diff
@@ -4246,9 +4246,9 @@ def _get_surf_area(comp):
 
             # 2.2 disables support for boosting.  The boosting parameter in 2.2 only has 'none' as an option, but
             # importing a bundle from old releases may still have 'linear' as an option, so we'll check here
-            if compute_kind in ['phoebe'] and self.get_value(qualifier='boosting_method', compute=compute, boosting_method=kwargs.get('boosting_method', None), **_skip_filter_checks) != 'none':
+            if compute_kind in ['phoebe'] and self.get_value(qualifier='boosting_method', compute=compute, boosting_method=kwargs.get('boosting_method', None), **_skip_filter_checks) not in ['none','manual']:
```
kecnry, line 4249: Since we're removing the parameter from the compute options, we don't need this to handle 'manual', but do need to not crash on the missing parameter.

```suggestion
            if compute_kind in ['phoebe'] and self.get_value(qualifier='boosting_method', compute=compute, boosting_method=kwargs.get('boosting_method', None), default='none', **_skip_filter_checks) != 'none':
```

phoebe/frontend/bundle.py
```diff
@@ -4246,9 +4246,9 @@ def _get_surf_area(comp):
...
                 report.add_item(self,
-                                "support for beaming/boosting has been removed from PHOEBE 2.2.  Set boosting_method to 'none'.",
+                                "support for interpolated ('linear') beaming/boosting has been removed since PHOEBE 2.2.  Set boosting_method to 'none' or 'manual'.",
```
kecnry, line 4251: (or something similar... this is hard to word concisely)

```suggestion
                                "support for interpolated ('linear') beaming/boosting in compute options has been removed since PHOEBE 2.2.  'manual' boosting is now supported in the dataset options instead.",
```

kecnry (commented):
Added a few more small suggestions regarding the checks for the old parameter.  Also feel free to just delete the old lines that you've commented out... I don't think we'll go back to that old implementation and the history will still live here in the diff.

phoebe/parameters/dataset.py
```diff
@@ -197,6 +197,14 @@ def lc(syn=False, as_ps=True, is_lc=True, **kwargs):
...
                                        required_shape=[None],
                                        description='Limb darkening coefficients')]
+        params += [ChoiceParameter(qualifier='boosting_method', copy_for={'kind': ['star'], 'component': '*'}, component='_default',
+                                   value=kwargs.get('boosting_method', 'none'), choices=['none', 'manual'],
+                                   description='Method to use for Doppler boosting')]
+        params += [FloatArrayParameter(visible_if='boosting_method:manual', qualifier='boosting_index',
+                                       copy_for={'kind': ['star'], 'component': '*'}, component='_default',
+                                       value=kwargs.get('boosting_index', 1.0), default_unit=u.dimensionless_unscaled,
+                                       required_shape=[None],
+                                       description='Boosting index')]
```
kecnry, line 207: ah missed one, can you add entries to the docstring for these as well?

aprsa:
So does this PR go into 2.5 or are you proposing to release it as a bugfix?

Raindogjones:
I have no strong feelings on how/when this should be released, but it does seem like a pretty big change which would be more consistent with a new version rather than a bugfix.

phoebe/backend/universe.py
```diff
@@ -1294,6 +1292,10 @@ def from_bundle(cls, b, component, compute=None,
         ld_coeffs['bol'] = b.get_value(qualifier='ld_coeffs_bol', component=component, context='component', ld_coeffs_bol=ld_coeffs_bol_override, **_skip_filter_checks)
         profile_rest_override = kwargs.pop('profile_rest', None)
         lp_profile_rest = {ds: b.get_value(qualifier='profile_rest', dataset=ds, unit=u.nm, profile_rest=profile_rest_override, **_skip_filter_checks) for ds in datasets_lp}
+        boosting_method_override = kwargs.pop('boosting_method', None)
+        boosting_method = {ds: b.get_value(qualifier='boosting_method', dataset=ds, component=component, ld_mode=boosting_method_override, **_skip_filter_checks) for ds in datasets_intens}
```
kecnry, line 1296:
```suggestion
        boosting_method = {ds: b.get_value(qualifier='boosting_method', dataset=ds, component=component, boosting_method=boosting_method_override, **_skip_filter_checks) for ds in datasets_intens}
```

phoebe/backend/universe.py
```diff
@@ -1294,6 +1292,10 @@ def from_bundle(cls, b, component, compute=None,
...
+        boosting_index_override = kwargs.pop('boosting_index', None)
+        boosting_index = {ds: b.get_value(qualifier='boosting_index', dataset=ds, component=component, ld_mode=boosting_index_override, **_skip_filter_checks) for ds in datasets_intens}
```
kecnry, line 1298:
```suggestion
        boosting_index = {ds: b.get_value(qualifier='boosting_index', dataset=ds, component=component, boosting_index=boosting_index_override, **_skip_filter_checks) for ds in datasets_intens}
```

Raindogjones:
> Testing shows that at run_compute() the boosting_method isn't getting picked up and is always 'none' even if it is set to 'manual'. I suspect this might be in `class Star(Body)` where I have `self.boosting_method = kwargs.get('boosting_method', 'none')`, kwargs.get isn't picking up the boosting_method so it is defaulting to 'none'. Bit stuck how to fix this though, if that even is the cause...

Seems to be working now following some pointers from @kecnry !

Raindogjones:
The failures here look (to me at least) to be a consequence of moving boosting from per-compute to per-dataset. I think every one of the fails is upon import of an old bundle that has boosting_method@compute set, so when loading it can't find the relevant twig. Previously, we allowed importing an old bundle with boosting_method@compute and then threw an error at run_compute() to say that it must be set to None. I'm not sure what the best way to handle this is now: 
Should the error be thrown at load() forcing boosting_method@dataset to be none and advising that if the users want boosting they need to set the bindex for those datasets? Thoughts @aprsa @kecnry ?

phoebe/parameters/dataset.py
```diff
@@ -197,6 +200,14 @@ def lc(syn=False, as_ps=True, is_lc=True, **kwargs):
...
```
kecnry, line 210: probably want to define limits on this to prevent going out of physical range

Raindogjones:
Tests are still going to fail and I'm not 100% sure why.  The problem is with the backwards compatibility test loading from v2.3. For some reason, the bundle loads with no 'boosting_method' at all (not per compute or per dataset).

kecnry:
@Raindogjones - can you just look through the few commits I added and make sure they look reasonable to you and then I think we can mark this as ready for review.  I'd love to get #847 in first and merged into here as well so that we can test this against 2.4 backwards compatibility as well (although I don't expect a problem there since it works now for 2.3).

Raindogjones (commented):
Everything looks good, I think. Thanks!

phoebe/frontend/bundle.py
```diff
@@ -916,7 +916,7 @@ def existing_value(param):
             # call set_hierarchy to force mass constraints to be rebuilt
             b.set_hierarchy()
 
-        if phoebe_version_import < StrictVersion("2.4.0") or ".dev" in version:
+        if phoebe_version_import < StrictVersion("2.4.0"):
             warning = "importing from an older version ({}) of PHOEBE to PHOEBE 2.4.  This may take some time.  Please check all values.".format(phoebe_version_import)
```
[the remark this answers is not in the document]
kecnry: that's a good point - the 2.4 was intentional (since it is updating to the requirements introduced in 2.4), but I just went through and changed these all to be consistent and say "to PHOEBE 2.x+" instead.

phoebe/backend/universe.py
```diff
@@ -1877,19 +1886,24 @@ def _populate_lc(self, dataset, ignore_effects=False, **kwargs):
...
+            # elif boosting_method == 'linear':
+            #     logger.debug("calling pb.bindex for boosting_method='linear'")
+            #     bindex = pb.bindex(Teff=self.mesh.teffs.for_computations,
+            #                        logg=self.mesh.loggs.for_computations,
+            #                        abun=self.mesh.abuns.for_computations,
+            #                        mu=abs(self.mesh.mus_for_computations),
+            #                        atm=atm,
+            #                        photon_weighted=intens_weighting=='photon')
+            #
+            #     boost_factors = 1.0 + bindex * self.mesh.velocities.for_computations[:,2]/37241.94167601236
```
kecnry, line 1906: Let's just get rid of this since it won't work anyways - it will always live in the git history if we need to revive it.

```suggestion
```

phoebe/frontend/default_bundles/default_binary.bundle
```diff
@@ -427,6 +427,7 @@ null
 "choices": [
 "auto",
 "phoenix",
+"tmap",
```
kecnry, line 430: tmap is not implemented here yet, is it?

```suggestion
```

phoebe/frontend/default_bundles/default_binary.bundle
```diff
@@ -775,6 +776,7 @@ null
...
```
kecnry, line 779:
```suggestion
```

phoebe/frontend/default_bundles/default_contact_binary.bundle
```diff
@@ -427,6 +427,7 @@ null
 "choices": [
 "auto",
 "phoenix",
+"tmap",
```
kecnry, line 430:
```suggestion
```

phoebe/frontend/default_bundles/default_contact_binary.bundle
```diff
@@ -775,6 +776,7 @@ null
...
```
kecnry, line 779:
```suggestion
```

phoebe/frontend/default_bundles/default_star.bundle
```diff
@@ -427,6 +427,7 @@ null
 "choices": [
 "auto",
 "phoenix",
+"tmap",
```
kecnry, line 430:
```suggestion
```

kecnry (commented):
Just a few remaining small tweaks and then I think this is ready to go in (with tutorial/docs updates as a follow-up before this gets released).  Thanks for all the hard work!

Raindogjones:
Agreed. The updates to the default bundles were actually from an old branch where I was working on tmap, I think. Shouldn't have crept in here, so good catch! In any case, they would have been fixed when the blending branch merges in.

kecnry (approved):

kecnry:
@aprsa - do you want to take a look at this before merge?

aprsa (approved):
looks good to me!

PR #739, opened 2023-05-23
Spot handling updates.

### LINKED ISSUE #716: rotstar spot support broken in 2.4.9
Reported by Marcin Wrona:

I have the newest version of the PHOEBE - 2.4.9. The bug concerns running a single star with a spot.

I wanted to run my own code but it kept failing, so I decided to test a code from the tutorial and I got the same result. I also tested it on the Web UI Client and got the same problem.

The code is crashing on b.run_compute() with an error "AttributeError: 'Star_rotstar' object has no attribute 'F'".

Here's the code from the tutorial:

```
import phoebe
from phoebe import u # units
import numpy as np

logger = phoebe.logger()

b = phoebe.default_star()

b.add_spot(radius=30, colat=80, long=0, relteff=0.9)
#print(b['spot'])

times = np.linspace(0, 10, 11)
b.set_value('period', 10)
b.add_dataset('mesh', times=times, columns=['teffs'])

b.run_compute(distortion_method='rotstar', irrad_method='none')
```

aprsa:
The inclusion of spots has been broken by a recent commit. Here we refactor some code to flag when the mesh needs to be populated and fix two bugs in the spot treatment. We also expand the spot test to include supersynchronous and misaligned cases. This does *not* fix misaligned spot issues, but it closes #716.


phoebe/backend/universe.py
```diff
@@ -3163,19 +3188,21 @@ def pointing_vector(self, s, time):
         """
         s is the spin vector in roche coordinates
         time is the current time
+        returns spot center in roche coordinates
         """
         t = time - self._t0
         longitude = self._longitude + self._dlongdt * t
+        print(f'{longitude=}')
```
kecnry, line 3195:
```suggestion
```

kecnry (commented):
In general this looks reasonable to me, but I would like to think through all the different `needs_recompute_instantaneous` cases that have changed to see if I can think of any scenario where those assumptions might cause troubles (since I've screwed that up in past bugfix attempts relating to spots)

PR #741, opened 2023-05-24
Fixes spots on single rotating stars.

### LINKED ISSUE #716: rotstar spot support broken in 2.4.9

aprsa:
Closes #716.

### DIFF
--- a/phoebe/backend/universe.py
+++ b/phoebe/backend/universe.py
@@ -2432,7 +2432,7 @@ class Star_rotstar(Star):
     def needs_recompute_instantaneous(self):
         # recompute instantaneous for asynchronous spots, even if meshing
         # doesn't need to be recomputed
-        return self.needs_remesh or len(self.features)
+        return self.needs_remesh or (not self.is_single and len(self.features) and self.F != 1)
 
     @property
     def needs_remesh(self):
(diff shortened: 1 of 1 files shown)


kecnry:
This is being over-agressive and will slow down the rotstar binary case, right?  Can we use anything stored in the star (`ind_sibling` perhaps - without introducing a lot of new code/logic) to determine if its a single star?

kecnry (approved):
Looks good to me as soon as CI passes - thanks!

PR #746, opened 2023-06-07
fix constraint migration for 2.3 -> 2.4

kecnry:
this fixes the migration when importing a bundle created in 2.3 into 2.4 that contains custom constraints.


aprsa (approved):
Looks good to me!

aprsa:
Do we have coverage for this in the test suite?

PR #784, opened 2023-09-28
Interferometric module for Phoebe

miroslavbroz:
Dear Andrej and Kyle, I'm not 100-% sure if I do this correctly, nevertheless, this is it. Clear skies, Mira


tests/nosetests/test_interferometry2/test_interferometry.py
```diff
@@ -0,0 +1,50 @@
...
+import numpy as np
+import phoebe
+from astropy import units
+
+logger = phoebe.logger(clevel='INFO')
+#logger = phoebe.logger(clevel='DEBUG')
+
+b = phoebe.default_binary()
+
+times, u, v, wavelengths, vises, sigmas = np.loadtxt("Vis.dat", usecols=[0, 1, 2, 3, 5, 6], unpack=True)
```
kecnry, line 13: I think this is failing in the CI because it needs an absolute path.  Something like the following should work (the other test will need similar updates).

```suggestion
import os

dir = os.path.dirname(os.path.realpath(__file__))

logger = phoebe.logger(clevel='INFO')
#logger = phoebe.logger(clevel='DEBUG')

b = phoebe.default_binary()

path = os.path.join(dir, "Vis.dat")
times, u, v, wavelengths, vises, sigmas = np.loadtxt(path, usecols=[0, 1, 2, 3, 5, 6], unpack=True)
```

miroslavbroz: Corrected. But 3 (unrelated) tests tend to fail...

miroslavbroz: Hmm, it has something to do with .github/workfows, gfortran.

phoebe/backend/backends.py
```diff
@@ -1,3 +1,5 @@
+
+import sys
```
kecnry, line 2: This doesn't seem to actually be used.

```suggestion
```

phoebe/backend/backends.py
```diff
@@ -297,14 +301,16 @@ def _extract_from_bundle(b, compute, dataset=None, times=None,
                     info['mesh_kinds'] = [b.filter(dataset=ds, context='dataset', **_skip_filter_checks).kind for ds in info['mesh_datasets']]
 
                 if by_time:
-                    for time_ in this_times:
+                    for i, time_ in enumerate(this_times):
+                        info['original_index'] = i
+                        info_ = info.copy()
```
kecnry, line 306: this still concerns me a little - does this work well with plotting and animating or would we need to update that logic as well?  If we do decide to go in this direction, we probably should change the implementation for line profiles as well 🤔

phoebe/backend/backends.py
```diff
@@ -1151,16 +1161,47 @@ def _run_single_time(self, b, i, time, infolist, **kwargs):
...
         packetlist = []
+        previous = None
         for k, info in enumerate(infolist):
             packet = dict()
 
-            # i, time, info['kind'], info['component'], info['dataset']
+            #i, time, info['kind'], info['component'], info['dataset']
             cind = starrefs.index(info['component']) if info['component'] in starrefs else None
-            # ts[i], xs[cind][i], ys[cind][i], zs[cind][i], vxs[cind][i], vys[cind][i], vzs[cind][i]
+            #ts[i], xs[cind][i], ys[cind][i], zs[cind][i], vxs[cind][i], vys[cind][i], vzs[cind][i]
```
kecnry, line 1179: what is changed here, just removing the space?  We don't (yet) adhere to codestyle checks, but I think the space after the comment is standard.

phoebe/backend/backends.py
```diff
@@ -1151,16 +1161,47 @@ def _run_single_time(self, b, i, time, infolist, **kwargs):
...
             kind = info['kind']
+            dataset = info['dataset']
+
+            # save baselines and wavelengths (for interferometry)
+            if kind == 'vis' and dataset != previous:
```
kecnry, line 1184: is the need for `previous` because of the "flattening" and using `original_index`?  Is there any concern if the input is not sorted correctly?

phoebe/backend/backends.py
```diff
@@ -1151,16 +1161,47 @@ def _run_single_time(self, b, i, time, infolist, **kwargs):
...
+                ucoord = b.get_value('u@'+dataset+'@dataset')
```
kecnry, line 1185: nothing internal should use twigs (`@`) as they add extra expense and ambiguity.  Instead, please explicity use each tag (`b.get_value(qualifier=u, dataset=dataset, context='dataset')`)

phoebe/backend/backends.py
```diff
@@ -1151,16 +1161,47 @@ def _run_single_time(self, b, i, time, infolist, **kwargs):
...
+                vcoord = b.get_value('v@'+dataset+'@dataset')
+                wavelengths = b.get_value('wavelengths@'+dataset+'@dataset')
+                previous = dataset
+
+                if_method = b.get_value('if_method@'+dataset+'@dataset')
```
kecnry, line 1190: this will need updating to accept kwarg overrides during `run_compute` (I can help with that - it probably isn't simple to figure out)

phoebe/backend/interferometry.py
```diff
@@ -0,0 +1,266 @@
...
+interferometry.py
+
+Interferometric module for Phoebe.
+
+Reference: Hanbury-Brown et al. (1974, MNRAS 167, 121-136).
+Reference: Brož (2017, ApJS 230, 19).
+
+"""
+
+import sys
```
kecnry, line 13: This is only used in a commented debug line.  If those commented lines should stay, then maybe we can move the import into that block?

```suggestion
```

phoebe/backend/universe.py
```diff
@@ -1762,17 +1798,21 @@ def _populate_lc(self, dataset, ignore_effects=False, **kwargs):
...
+
         lc_method = kwargs.get('lc_method', 'numerical')  # TODO: make sure this is actually passed
 
         passband = kwargs.get('passband', self.passband.get(dataset, None))
         intens_weighting = kwargs.get('intens_weighting', self.intens_weighting.get(dataset, None))
         atm = kwargs.get('atm', self.atm)
         extinct = kwargs.get('extinct', self.extinct)
         Rv = kwargs.get('Rv', self.Rv)
-        ld_mode = kwargs.get('ld_mode', self.ld_mode.get(dataset, None))
+        ld_mode = kwargs.get('ld_mode', self.ld_mode.get(dataset, 'manual'))
```
kecnry, line 1811: why this change?  Is this a bugfix or does it introduce a change in behavior?

phoebe/backend/universe.py
```diff
@@ -301,7 +307,7 @@ def update_positions(self, time, xs, ys, zs, vxs, vys, vzs,
                                  ds=ds, Fs=Fs, ignore_effects=ignore_effects)
 
 
-    def populate_observables(self, time, kinds, datasets, ignore_effects=False):
+    def populate_observables(self, time, kinds, datasets, ignore_effects=False, **kwargs):
```
kecnry, line 310: is this (`**kwargs` here and in calls to `populate_observable`) used in this PR?

phoebe/backend/universe.py
```diff
@@ -508,7 +514,7 @@ def handle_eclipses(self, expose_horizon=False, **kwargs):
         return horizon
 
 
-    def observe(self, dataset, kind, components=None, **kwargs):
+    def observe(self, dataset, kind, components=None, info=None, **kwargs):
```
kecnry, line 517: this change doesn't seem to be used

phoebe/backend/universe.py
```diff
@@ -631,7 +637,9 @@ def gaussian(sv):
         else:
             raise NotImplementedError("observe for dataset with kind '{}' not implemented".format(kind))
 
+        # Note: See interferometry.vis_integrate().
 
+        # Note: See interferometry.clo_integrate().
```
kecnry, line 642:
```suggestion
```

phoebe/parameters/dataset.py
```diff
@@ -778,8 +777,86 @@ def mesh(syn=False, as_ps=True, **kwargs):
 
     return ParameterSet(params) if as_ps else params, constraints
 
+def vis(syn=False, as_ps=True, **kwargs):
+    """
+    Create a <phoebe.parameters.ParameterSet> for an interferometric visibility |V|^2 dataset.
+
+    Note: See lc().
+
```
kecnry, line 785: these will need full docstrings since they are used to generate the API docs on the website and all new qualifiers will need to be added to `forbidden_labels` (defined in `parameters.py`, looks like you already added at least some or maybe even all of them) and these methods included in `test_forbidden_parameters`.  This prevents the user from naming a dataset something that conflicts with a qualifier and would result in ambiguous filtering when using twigs.

phoebe/parameters/parameters.py
```diff
@@ -3647,6 +3648,8 @@ def calculate_residuals(self, model=None, dataset=None, component=None,
         Currently supports the following datasets:
         * <phoebe.parameters.dataset.lc>
         * <phoebe.parameters.dataset.rv>
+        * <phoebe.parameters.dataset.vis>
+        * <phoebe.parameters.dataset.clo>
```
kecnry, line 3652: what about t3?

kecnry (commented):
Overall this look great!  Hopefully this won't cause too many conflicts with other efforts since most of the code is isolated to new functionality.  Here are a few comments from a first-pass (still need to actually test the implementation):

* I think we need to debate/decide what to name these new dataset types.  `vis` to me can be ambiguous and I fear would be confusing with mesh visibilities in general.  Maybe we even want to have some sort of prefix for interferometry if we need 3 different dataset types.
* As mentioned in the in-line comment, I think the `original_index` idea needs some serious thought and testing before we commit to it.  I will give that some thought when I get a chance to actually pull this and test it.  It is a significant change to users though if we decide to change line profiles and we need to think about whether the convenience on our end is worth the possible inconvenience to them.  It is possible we might be able to make filtering smarter to automatically slice the arrays correctly though and still filter by time even without using the dedicated time tag.
* We'll probably want to try to minimize the data files included just for tests to minimal examples to avoid (further) bloating the full package size.

phoebe/backend/backends.py
miroslavbroz: Yes and no -- the plotting works, but a plot 'vises' vs. 'u' (and line style) is not sorted according to 'u', so it ends up with criss-crossing lines. Do you know why, please? A gnuplot script this plots correctly...

phoebe/backend/backends.py
miroslavbroz: Corrected.

phoebe/backend/backends.py
miroslavbroz: Corrected.

phoebe/backend/backends.py
miroslavbroz: Here, it is useless to save baselines and wavelengths in each cycle (over k, info). It is an useful optimization. Btw. the b.get_value() calls are pretty expensive, especially if present in the innermost cycle!

phoebe/backend/backends.py
miroslavbroz: Corrected.

phoebe/backend/universe.py
miroslavbroz: Corrected. It was in a version with observe(); obsolete.

phoebe/backend/universe.py
miroslavbroz: Corrected. Ditto.

phoebe/backend/universe.py
miroslavbroz: Corrected.

phoebe/parameters/dataset.py
miroslavbroz: Corrected. Copy & paste.

phoebe/parameters/parameters.py
miroslavbroz: Corrected.

phoebe/backend/interferometry.py
miroslavbroz: Corrected.

phoebe/backend/universe.py
miroslavbroz: Corrected. It was in a version without lc_params; obsolete.

phoebe/backend/backends.py
miroslavbroz: Not yet done. I prefer to change nothing else (compute.py).

phoebe/backend/backends.py
kecnry: Yes, `get_value` calls should be minimized as much as possible (by pulling any values in advance that need to be accessed multiple times, etc).  Since you're pulling multiple values from the dataset, you can also save quite a bit of time by skipping multiple visibility checks and by pre-filtering to a smaller common set of parameters so that future calls have less parameters to search through.  For example:

```
ds_ps = b.filter(dataset=dataset, context='dataset', **_skip_filter_checks)
...
ucoord = ds_ps.get_value(qualifier='u', **_skip_filter_checks)
vcoord = ds_ps.get_value(qualifier='v', **_skip_filter_checks)
```

kecnry (commented):
Ok, let's try to just wrap up a few things and then I think we can merge this and defer the decision about `original_index` (I don't want to keep holding this up on that, but still think we need to give it careful thought and testing).

* add `_**skip_filter_checks` to all internal filter calls (https://github.com/phoebe-project/phoebe2/pull/784/files#r1350245300)
* are all of the added files actually used either in the code or tests?  If not, let's remove them for now so they don't pollute the git history (we can always add them later if we need them).  
* can the tests have any checks/asserts?  Right now they just seem to run and export to external files, so I guess we are testing that they don't crash, but aren't testing the actual results at all.  We can also defer this if we need to.
* test logs are easier to parse if they don't print anything by default.  We usually do this by having the test in a function with `verbose=False` by default, and then call that function with `verbose=True` from `if __name__ == '__main__'`.

miroslavbroz:
In interferometry.py, '_**skip_filter_checks' is not used, elsewhere it is used consistently. All files are used. I added assert() to tests. Now, they do not print.

miroslavbroz:
Oh, no! Some tests are failing -on specific system configurations- which is probably a problem with single vs. double precision computations? Shall I then decrease the precision in assert() to 10^-6? Hmm...

kecnry:
> In interferometry.py, '_**skip_filter_checks' is not used, elsewhere it is used consistently. 

Is there a reason for not using it?  We're paying a lot of extra overhead.

> All files are used. 

Is there any way we can achieve the same test coverage but without all the large data files?  This diff is currently almost 17k lines...  We can always minimize them later, but then that will always be in the git history, which is why I'm trying to see if that is all necessary to have in the repo or not.

> Shall I then decrease the precision in assert() to 10^-6?

1E-5 (in your latest commit) seems concerning to me - and seems like more than just computational noise.  But we can look into that further as a follow-up, if necessary.

miroslavbroz:
> > In interferometry.py, '_**skip_filter_checks' is not used, elsewhere it is used consistently.
> 
> Is there a reason for not using it? We're paying a lot of extra overhead.

There is ZERO b.filter() or b.get_value() call in interferometry.py. Elsewhere, e.g., in backend.py, I looked at the code and I used '_**skip_filter_checks' consistently. Do you want me to add it elsewhere? If I use it here:

'''
                ucoord = b.get_value(qualifier='u', dataset=dataset, context='dataset', **_skip_filter_checks)
'''

Why it is not used here?

'''
                wavelengths = b.get_value(qualifier='wavelengths',
                                          component=info['component'],
                                          dataset=info['dataset'],
                                          context='dataset',
                                          unit=u.nm)
'''

kecnry (approved):
Ok, let's get this in for now to start the branch.  I made notes for the items that we need to revisit in the future.  Thanks for your patience!

PR #785, opened 2023-10-01
Numpy array comparison fix

aprsa:
Old versions of numpy could compare different length arrays element-wise (i.e., np.all(x==y) for x.shape != y.shape). That is no longer the case for numpy 1.24 onwards. Note that the fix uses len(x) instead of x.shape comparison to be more general, i.e. it works also if x and y are other types of iterables, not just ndarrays.


kecnry (approved):
Simple enough - thanks!

PR #786, opened 2023-10-13
Simplified CI to account for most (all) practical use cases

aprsa:
The CI will no longer test contrived combinations of dependencies; instead, for each python version (3.7 through 3.11) and each architecture (ubuntu-latest and macos-latest) it will pull the latest compatible dependencies from pip and build with those. This way any user would be able to recreate by spinning a venv and pip installing phoebe for their version of the python interpreter. It also helps us catch any issues with upstream dependencies as they arise so that we can limit versions when we need to.


kecnry:
Is there anything here that wasn't included in #787?  If not, let's just close without merging.  Otherwise - we'll need to rebase and hope that all tests pass.

aprsa:
We're good with #787; I had to merge in #786 into #787 to get the tests to pass. Closing now.

PR #787, opened 2023-10-14
Fixes the spuriously failing "test_rm" test.

### LINKED ISSUE #689: Spurious failure in test_rm
Test_rm.py:test_binary is failing; this combination of dependencies is currently excluded from the CI matrix so it is very urgent to get this fixed and remove it from exclusions.

aprsa:
This was due to the numerical instability that was caused by computing RVs precisely at conjunctions of precisely identical stars except for sub/super-synchronous rotation. Fixed by making the secondary slightly smaller and, for good measure, slightly less massive. Closes #689.


kecnry:
It looks like this includes #786 but is also needed in order for the tests to pass, is that right?  Should we just handle them both in this single PR then and close #786?

aprsa:
I'd just do both at the same time, because that opens the path for [nodistutils](https://github.com/Raindogjones/phoebe2/tree/nodistutils) to be merged in.

tests/nosetests/test_rm/test_rm.py
```diff
@@ -1,52 +1,60 @@
...
     else:
         b.import_model(os.path.join(os.path.dirname(__file__), 'test_rm_{}.comp.model'.format(syncpar)), model='legnumresults', overwrite=True)
 
+    phoebe1_rv1 = b.get_value('rvs@primary@legnumresults@legnum')
+    phoebe1_rv2 = b.get_value('rvs@secondary@legnumresults@legnum')
+
+    phoebe2_rv1 = b.get_value('rvs@primary@phnumresults@phnum')
+    phoebe2_rv2 = b.get_value('rvs@secondary@phnumresults@phnum')
+
+    phoebe1_rv2[np.isnan(phoebe2_rv2)] = np.nan
```
kecnry, line 21: what does phoebe1 put in these locations instead?  (mostly just curious)

kecnry (approved):
Sorry it took a while to get to this - this all looks reasonable to me for now.  We may eventually want try to still get predictable results in these cases, but agree that this is a more sensible test and its better to have this test running and passing than chasing down small numerical errors in an edge case.  Thanks!

tests/nosetests/test_rm/test_rm.py
aprsa: It uses the dynamical values as a drop-in replacement.

PR #788, opened 2023-10-31
Replacing the python interpolation prototype with the C implementation

aprsa:
This is a massive commit that syncs upstream with the ported interpolator into C. It remains work in progress, but at this point all the tests are passing and the interpolation functionality is fully restored (and optimized). Extrapolation and blending are currently offline. Also, there may be memory leaks in the interpolator that still need addressing.


aprsa:
Superseded by [this PR](https://github.com/phoebe-project/phoebe2/pull/791).

PR #792, opened 2023-11-07
Interpolation rewrite into C

aprsa:
This is the third and hopefully the final attempt to get a clean PR for this large change. It incorporates the test_rm fix without subsequent blunders present in blending-dev but otherwise attempts to keep code changes outside of passbands.py and C implementation minimal. All tests are passing on the local repo so all tests should be passing here as well. Note that the fortran CI fix is not incorporated here, so some tests might be spuriously failing.


phoebe/frontend/bundle.py
```diff
@@ -10817,7 +10819,7 @@ def compute_pblums(self, compute=None, model=None, pblum=True, pblum_abs=False,
 
                     logger.info("estimating pblum for {}@{} using atm='{}' and stefan-boltzmann approximation".format(dataset, component, atm))
                     # requiv in m, Inorm in W/m**3, ldint unitless, ptfarea in m -> pblum_abs in W
-                    pblums_abs[dataset][component] = 4 * np.pi * requivs[component]**2 * abs_normal_intensities * ldint * ptfarea
+                    pblums_abs[dataset][component] = float(4 * np.pi * requivs[component]**2 * abs_normal_intensities * ldint * ptfarea)
```
kecnry, line 10822: which of these are lists/arrays (I think `abs_normal_intensities` and `ldint` since those are being queried from the backend, right)?  Is there a performance difference between casting those directly, taking index 0, vs casting to float at the end?

phoebe/frontend/default_bundles/default_contact_binary.bundle
```diff
@@ -2871,7 +2866,7 @@ null
 "qualifier": "phoebe_version",
 "context": "setting",
 "description": "Version of PHOEBE",
-"value": "2.4.8.dev+feature-blending",
+"value": "2.4.3.dev+feature-blending",
```
kecnry, line 2869: this looks like a reversion - can you recreate the default bundles just to make sure they're up to date with the new/revised blending parameters?  And we'll have to decide whether tmap does or does not belong in the list of atmospheres at this point.

phoebe/parameters/compute.py
```diff
@@ -300,6 +300,9 @@ def legacy(**kwargs):
...
     params += [IntParameter(visible_if='fti_method:oversample', qualifier='fti_oversample', copy_for={'kind': ['lc'], 'dataset': '*'}, dataset='_default', value=kwargs.get('fti_oversample', 5), limits=(1,None), default_unit=u.dimensionless_unscaled, description='Number of times to sample per-datapoint for finite-time integration')]
 
+    params += [ChoiceParameter(visible_if='atm:!blackbody', copy_for = {'kind': ['star'], 'component': '*'}, component='_default', qualifier='blending_method', value=kwargs.get('blending_method', 'linear'), choices=['none', 'nearest', 'linear'], description='Method to use for blending')]
+    params += [ChoiceParameter(visible_if='atm:!blackbody,blending_method:!none', copy_for = {'kind': ['star'], 'component': '*'}, component='_default', qualifier='ld_blending_method', value=kwargs.get('ld_blending_method', 'nearest'), choices=['none', 'nearest', 'linear'], description='Method to use for extrapolating limb-darkening during blending (for all datasets and bolometric for irradiation, if applicable)')]
+
```
kecnry, line 305: it looks like these already exist in the docstrings - can you just confirm that they're up to date with any changes here?

tests/nosetests/test_blackbody/test_blackbody.py
```diff
@@ -31,14 +31,16 @@ def test_binary(plot=False, gen_comp=False):
...
 
-    #turn off albedos (legacy requirement)
+    # turn off albedos (legacy requirement)
     b.set_value_all('irrad_frac_refl_bol',  0.0)
 
+    b.set_value_all('blending_method', 'none')
+
     if plot: print("running phoebe2 model...")
-    b.run_compute(compute='phoebe2', irrad_method='none', model='phoebe2model')
+    b.run_compute(compute='phoebe2', irrad_method='none', model='phoebe2model', blending_method='none')
```
kecnry, line 40: these are redundant, but shouldn't hurt either

phoebe/dependencies/ndpolator/ndpolator.py
```diff
@@ -13,13 +39,13 @@ def __init__(self, axes, grid, impute=False):
...
             raise NotImplementedError('consider if this is worth implementing here.')
-        
+
         self.indices = np.argwhere(~np.isnan(grid[...,0]))
         non_nan_vertices = np.array([ [self.axes[i][self.indices[k][i]] for i in range(len(self.axes))] for k in range(len(self.indices))])
         self.nntree = cKDTree(non_nan_vertices, copy_data=True)
         ranges = (range(len(ax)-1) for ax in axes)
         slices = [tuple([slice(elem, elem+2) for elem in x]) for x in product(*ranges)]
-        self.ics = np.array([s.start for slc in slices for s in slc if ~np.any(np.isnan(grid[slc]))], dtype=int).reshape(-1, len(axes))
+        self.ics = np.array([s.start for slc in slices for s in slc if ~np.any(np.isnan(grid[slc]))], dtype=int).reshape(-1, len(axes))  # bottleneck
```
kecnry, line 48: how much of a bottleneck?  Should we open a ticket so that we remember to get back to this someday?

kecnry (commented):
Overall looks good and will be nice to finally be able to test this!  A lot of minor changes could arguably be moved or copied to the main dev branch, but we can deal with that later if they're too difficult to move now.

phoebe/frontend/bundle.py
aprsa: yep, both `abs_normal_intensities` and `ldint` are now always arrays. I can't imagine there is any appreciable difference between those scenarios given a plethora of other bottlenecks. :)

phoebe/parameters/compute.py
aprsa: I did add both of these and I did not touch the doctring, so I assume that this is all correct. (this is for an alternate backend, not phoebe)

tests/nosetests/test_blackbody/test_blackbody.py
aprsa: I was overwriting the value while testing, and this work continues, that's why I left it in there.

phoebe/dependencies/ndpolator/ndpolator.py
aprsa: this is being ported into C right now.

phoebe/frontend/default_bundles/default_contact_binary.bundle
aprsa: recreated, tests running.

phoebe/parameters/compute.py
kecnry: Ok, this is for legacy - can you please add the entries to the docstring for that function as well then (and any other backends where it is added).  You can copy-paste from the phoebe backend docstring.

tests/nosetests/test_blackbody/test_blackbody.py
kecnry: Does it work with and without line 37?  It might be nice to test a case each way to make sure that the backend is accepting the in-line kwarg override as expected.

aprsa: I'm inclined to defer that test to the time when extrapolation is fully ported to C.

kecnry (approved):
Thanks!

PR #793, opened 2023-11-09
Dropping distutils dependencies across the code

aprsa:
In addition to the separate PR that removes distutils from setup.py, there were a number of distutils' LooseVersion() and StrictVersion() uses throughout the code. These have been replaced by packaging.version.parse. In addition, the code that added IAU 2015 resolution units to astropy <= 2.0 has been removed as we require astropy 3+.


kecnry (approved):
Looks good to me - thanks!

Would you mind also making a PR directly to autofig and distl with these changes so they don't get reintroduced here later?  (But since we're not releasing those very often, I think its find to just make the changes directly here for now since they shouldn't actually affect any behavior).

PR #796, opened 2023-11-20
Build update

aprsa:
This is an upgrade from @Raindogjones' previous PR to replace distutils; this one goes farther by switching to pyproject.toml, using setuptools as backend and build/pip as frontend.

Note that C/C++ compiler versions/flags have not been ported (yet); the only currently supported way to do that is clunky (it goes back to setup.py with custom functions) but there's a [PEP 725](https://discuss.python.org/t/pep-725-specifying-external-dependencies-in-pyproject-toml/31888?page=1) under consideration that would streamline this significantly. @horvatm should weigh in on this whether we should re-add the tests into setup.py or await PEP 725.

Closes #644


Raindogjones:
As of right now, the C/C++ stuff looks like a show-stopper for my system.
`      clang: error: linker command failed with exit code 1 (use -v to see invocation)`
`      error: command '/usr/bin/clang++' failed with exit code 1`
`      [end of output]`
`note: This error originates from a subprocess, and is likely not a problem with pip.`
`ERROR: Failed building wheel for phoebe`

aprsa:
Can you please share your system details?

Raindogjones:
Tried both with and without -e, still fails (albeit with a different error: using -e tells me to try not using -e).

I'm on OS X 13.6.1 on a 2018 intel Mac
Apple clang version 15.0.0 (clang-1500.0.40.1)
gcc version 12.3.0 (Homebrew GCC 12.3.0) 

I'm a bit confused about the error it throws (that could be because it's late and I'm tired though). Looks to me like it should be gcc-12 but then complains about a clang error - see attached
[output.txt](https://github.com/phoebe-project/phoebe2/files/13420237/output.txt)

aprsa:
The latest macos [runner](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) that github actions support is macos-12. Thus, I don't know how to test/troubleshoot macos-13 without having my hands on one. Does building sdist/wheel work? `python -m build`, probably after `pip install build`.

Raindogjones:
Nope, looks a lot like the same error.
[output.txt](https://github.com/phoebe-project/phoebe2/files/13421444/output.txt)

aprsa:
I think it might be [your system](https://developer.apple.com/forums/thread/737707), not phoebe. Can you link any c/c++ executables?

aprsa:
A possible `CXXFLAGS='-Wl,-ld_classic'` workaround is [mentioned](https://github.com/Homebrew/homebrew-core/issues/145991); does that help?

Raindogjones:
Yeah, this is weird. Can't link with g++-12 or g++13 but clang++ is working. So setting CC=clang worked. Problem was definitely on my end.

horvatm (approved):
I don't see anything problematic.

kecnry:
> So setting CC=clang worked.

@Raindogjones - is this something that you didn't need to do manually before though?  And if it is just a weird setup, should we provide some advice somewhere on the install page for how to handle these cases?

When we release this, I can also update the install page to suggest `pip install -e .` in the [Installing from Source](http://phoebe-project.org/install#source) instructions.  Anything else I should add either there or elsewhere on that page?

kecnry (commented):
Is all the renaming from "nosetests" to "tests" to actually migrate to pytest or is it required for the rest of the changes?

I am just worried about dealing with conflicts in feature branches as we try to propagate this there.  If we don't need to do it now, maybe we add it to the list of changes to make once we can close out those feature branches?

Raindogjones:
@kecnry I don't think any of that is necessary.  Seems this is a common problem of apple's xcode 15 not playing well with homebrew's gcc.  I hadn't realised before because I only updated xcode ~a month ago and haven't done much coding since.  Easy workarounds are to build gcc from source or to stick to clang.  I also read somewhere that this is fixed in the current xcode beta so will go away soon.

Raindogjones (approved):
Long story short, I can now install easily with either clang or gcc on my mac. So, I see no reason to hold this.

tests/run_tests.py
```diff
@@ -16,7 +16,7 @@
 import subprocess
 
 if len(sys.argv)==1:
-    do = ['tutorials', 'doctests', 'nosetests', 'pylint', 'benchmark']
+    do = ['tutorials', 'doctests', 'tests', 'pylint', 'benchmark']
```
kecnry, line 19: this will also need a change to website instructions

aprsa:
> Is all the renaming from "nosetests" to "tests" to actually migrate to pytest or is it required for the rest of the changes?

It is not required for the build system switch, I just seized the opportunity since I was updating the CI action that runs the tests. Keeping it labeled as nosetests is confusing and might make people want to run nose instead of pytest.

kecnry (approved):

setup.py
```diff
@@ -1,459 +1,22 @@
...
-                   'phoebe.dependencies.ligeor',
-                   'phoebe.dependencies.ligeor.ebai', 'phoebe.dependencies.ligeor.ebai.database', 'phoebe.dependencies.ligeor.eclipse', 'phoebe.dependencies.ligeor.models', 'phoebe.dependencies.ligeor.utils'],
-       install_requires=['numpy>=1.15','scipy>=1.4','astropy>=3.0', 'corner', 'pytest', 'requests', 'tqdm', 'python-socketio[client]']+['flask', 'flask-cors', 'flask-socketio==4.3.*', 'gevent-websocket'],
-       package_data={'phoebe.atmospheres':['tables/wd/*', 'tables/passbands/*'],
-                     'phoebe.frontend':['default_bundles/*.bundle'],
-                     'phoebe.solverbackends.ebai': ['*.data', '*.weights'],
-                     'phoebe.solverbackends.knn': ['*.knn']
-                    },
-       ext_modules = ext_modules,
-       scripts=None if  _env_variable_bool('PHOEBE_SKIP_SCRIPTS', False) else ['client-server/phoebe-server', 'client-server/phoebe-autofig'],
```
kecnry, line 453: I missed this - but it [seems this removed support for installing these scripts](https://github.com/phoebe-project/phoebe2-ui/issues/26)... can this be reintroduced in the current setup?

PR #819, opened 2024-01-19
Import from within source tree overly restrictive

### LINKED ISSUE #806: Tests fail when phoebe is installed with `pip install -e .`
It says `You cannot import Phoebe from inside its main source tree.`

aprsa:
In order to avoid same-name module conflicts with the system module (such as io.py), phoebe shouldn't be imported from its own source tree. The original test eliminated the entire package tree, including all auxiliary directories, which is overly restrictive and it prevented pytest from running when phoebe was installed with `pip -e`. This PR fixes that by blocking imports from the root directory and from the root/phoebe tree (the actual sources). Closes #806.

### DIFF
--- a/phoebe/__init__.py
+++ b/phoebe/__init__.py
@@ -27,9 +27,8 @@ import atexit
 import re
 
 # People shouldn't import phoebe from the root directory or from root/phoebe:
-cwd = _os.getcwd()
-root_dir = _os.path.abspath(_os.path.split(_os.path.split(__file__)[0])[0])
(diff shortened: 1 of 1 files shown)


kecnry (approved):
Fine with me once CI is green!

aprsa:
Before we merge in though, it just occurred to me that this is in __init__; should we avoid defining phoebe.cwd and phoebe.root_dir, or at least prepend a '_'?

kecnry:
We really should just use `__all__`.... but we can do that later repo-wide

PR #820, opened 2024-01-20
update tests for changes to pytest

kecnry:
opened on behalf of @aprsa


tests/tests/test_backward_compatibility/test_import.py
```diff
@@ -101,7 +97,6 @@ def test_21(verbose=False, plot=False):
     if plot:
         b.plot(show=True, time=0)
 
-    return b
```
kecnry, line 104: are these intentionally removed (in several places throughout this PR)?  They can be useful when running the script directly and debugging the bundle.  If we do need to remove these, then the calls in `__main__` need to be updated as well.

tests/tests/test_misaligned/test_misaligned_area_volume.py
```diff
@@ -5,23 +5,23 @@
...
-  # detacted case (only meaningful)
-  choice = 0
-  q = 1
-  F = 1
-  d = 1
-  Omega0 = 10
-  theta = 0.9
-  area0 = 0.1554255703973858
-  volume0 = 5.7617852701434746e-03
+    # detacted case (only meaningful)
```
kecnry, line 10: while we're changing other things 😝 

```suggestion
    # detatched case (only meaningful)
```

kecnry (commented):
A little difficult to tease out the needed changes from everything else here, but I think everything looks reasonable.  Just one general concern about removing the bundle returns since that has been helpful to debug failing tests in the past.

tests/tests/test_backward_compatibility/test_import.py
kecnry: (turns out this is necessary otherwise pytest yells, so we'll just need to manually make these edits if we need to inspect the bundle to debug a broken test).

aprsa: Yep, pytest currently raises a warning but the next version will raise an error whenever tests return anything. So we can't put those back in, unfortunately. Here [here](https://github.com/pytest-dev/pytest/issues/7337) for details.

PR #823, opened 2024-01-22
Dynamical RVs now avoid meshing

### LINKED ISSUE #812: dynamical RVs should not need calls to atmosphere tables
```
import phoebe
b = phoebe.default_binary()
b.add_dataset('rv', compute_times=[0, 0.1])
b.set_value_all('rv_method', 'dynamical')
b.set_value_all('teff', 100000)
b.run_compute()
```

raises an error from atmosphere tables (or limb-darkening lookup) being out-of-bounds.... but so long as meshes are not required to be populated, or passband scaling to be used, PHOEBE should not need to make any calls to the atmosphere tables.

See https://github.com/phoebe-project/phoebe2/discussions/809#discussioncomment-7837270 for more details.

aprsa:
Calling b.compute_pblums() built the mesh and treated dynamical RVs as mesh-dependent. This fixes that for both run_compute() and for direct compute_pblums(), compute_l3(), and compute_ld_coeffs bundle methods. A new function, b._datasets_that_require_meshing(), filters datasets that require a mesh, i.e. 'lc', 'lp' and flux-weighted 'rv'. Closes #812.

### DIFF
--- a/phoebe/frontend/bundle.py
+++ b/phoebe/frontend/bundle.py
@@ -10389,7 +10389,7 @@ class Bundle(ParameterSet):
         return system
 
     def _datasets_where(self, compute, mesh_needed=False, l3_needed=False):
-        datasets = self.filter(compute=compute, qualifier='enabled', value=True, **_skip_filter_checks).datasets
+        datasets = self.filter(compute=compute, context='compute', qualifier='enabled', value=True, **_skip_filter_checks).datasets
         ds_kinds = [self.filter(dataset=ds, context='dataset', **_skip_filter_checks).kind for ds in datasets]
         backend = self.filter(compute=compute, context='compute', **_skip_filter_checks).kind
 
--- a/phoebe/frontend/bundle.py
+++ b/phoebe/frontend/bundle.py
@@ -10461,7 +10461,7 @@ class Bundle(ParameterSet):
             raise TypeError("compute must be a single value (string)")
 
         # either take user-passed datasets or datasets that have an l3_mode:
-        datasets = kwargs.pop('dataset', self._datasets_where(compute=compute, l3_needed=True))
+        datasets = kwargs.pop('dataset') if 'dataset' in kwargs else self._datasets_where(compute=compute, l3_needed=True)
         if isinstance(datasets, str):
             datasets = [datasets]
 
--- a/phoebe/frontend/bundle.py
+++ b/phoebe/frontend/bundle.py
@@ -10246,8 +10246,9 @@ class Bundle(ParameterSet):
             appropriate length given the respective value of `ld_func`).
         """
 
+        # check to make sure value of passed compute is valid
         if compute is None:
-            if len(self.computes)==1:
+            if len(self.computes) == 1:
                 compute = self.computes[0]
             else:
                 raise ValueError("must provide compute")
@@ -10258,7 +10259,9 @@ class Bundle(ParameterSet):
         # we'll add 'bol' to the list of default datasets... but only if bolometric is needed for irradiation
         needs_bol = compute_ps.get_value(qualifier='irrad_method', irrad_method=kwargs.get('irrad_method', None), default='none', **_skip_filter_checks) != 'none'
 
-        datasets = kwargs.pop('dataset', self.datasets + ['bol'] if needs_bol else self.datasets)
+        datasets = kwargs.pop('dataset', self.datasets + ['bol'] if needs_bol else self._datasets_that_require_meshing(compute))
+        if len(datasets) == 0:
+            return {}
         components = kwargs.pop('component', self.components)
 
         # don't allow things like model='mymodel', etc
@@ -10383,6 +10386,16 @@ class Bundle(ParameterSet):
 
         return system
 
+    def _datasets_that_require_meshing(self, compute):
+        print('entering _datasets_that_require_meshing()')
+        datasets = self.filter(compute=compute, qualifier='enabled', value=True, **_skip_filter_checks).datasets
+        ds_kinds = [self.filter(dataset=ds, context='dataset', **_skip_filter_checks).kind for ds in datasets]
+        return [ds for ds, kind in zip(datasets, ds_kinds) 
+                if kind == 'lc'
+                or kind == 'lp'
+                or (kind == 'rv' and len(self.filter(qualifier='rv_method', dataset=ds, compute=compute, value='flux-weighted', **_skip_filter_checks)) > 0)
+        ]
+
     def compute_l3s(self, compute=None, use_pbfluxes={},
                    set_value=False, **kwargs):
         """
@@ -10426,11 +10439,10 @@ class Bundle(ParameterSet):
         """
         logger.debug("b.compute_l3s")
 
+        # either take user-passed datasets or datasets that have an l3_mode:
         datasets = kwargs.pop('dataset', self.filter('l3_mode', check_visible=True).datasets)
         if isinstance(datasets, str):
             datasets = [datasets]
-
-
 
         if compute is None:
             if len(self.computes)==1:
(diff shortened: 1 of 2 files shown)


phoebe/frontend/bundle.py
```diff
@@ -11912,8 +11933,10 @@ def restore_conf():
                 # TODO: have this return a dictionary like pblums/l3s that we can pass on to the backend?
 
                 # we need to check both for enabled but also passed via dataset kwarg
-                ds_kinds_enabled = self.filter(dataset=dataset_this_compute, context='dataset', **_skip_filter_checks).kinds
-                if 'lc' in ds_kinds_enabled or 'rv' in ds_kinds_enabled or 'lp' in ds_kinds_enabled:
+                # ds_kinds_enabled = self.filter(dataset=dataset_this_compute, context='dataset', **_skip_filter_checks).kinds
+                datasets = self._datasets_that_require_meshing(compute)
```
kecnry, line 11937: looking at this closer, `compute_l3s` defaults to a slightly different list of datasets (those with `l3_mode` parameters).  Right now these are just `lc` datasets which also are always including in `_datasets_that_require_meshing`, but I'm a bit worried this could break in the future.  We could:
* add a comment that hopefully we'll see if we ever have any other dataset type use third light (or ever allow light curves without needing meshing, but still needing l3 logic)
* do something like `set(self._datasets_that_require_meshing(compute) + self._datasets_that_require_l3(compute))` 😬

tests/tests/test_rvs/test_dynrvs.py
```diff
@@ -0,0 +1,32 @@
+import phoebe
+import numpy as np
+
+
+def test_dynrvs(verbose=False):
+    """
+    This test checks whether PHOEBE computes dynamical RVs without building
+    a mesh, and whether the computed RVs are close to the flux-weighted RVs.
```
kecnry, line 8:
```suggestion
    This test checks whether PHOEBE computes dynamical RVs without building
    a mesh (as otherwise run_compute would fail with atm out-of-bounds errors), 
    and whether the computed RVs are close to the flux-weighted RVs.
```

phoebe/frontend/bundle.py
aprsa: I reworked the logic, see what you think in the commit that's about to hit the branch.

tests/tests/test_rvs/test_dynrvs.py
```diff
@@ -0,0 +1,33 @@
...
+    b.set_value('rv_method', component='primary', dataset='fluxrv', value='flux-weighted')
+    b.set_value('rv_method', component='secondary', dataset='fluxrv', value='flux-weighted')
+    b.run_compute(eclipse_method='only_horizon')
+
+    if verbose:
+        print('maximum primary RV offset:', np.abs(b['value@rvs@primary@dynrv']-b['value@rvs@primary@fluxrv']).max())
+        print('maximum secondary RV offset:', np.abs(b['value@rvs@secondary@dynrv']-b['value@rvs@secondary@fluxrv']).max())
+
+    assert np.allclose(b['value@rvs@primary@dynrv'], b['value@rvs@primary@fluxrv'], atol=0.3)
+    assert np.allclose(b['value@rvs@secondary@dynrv'], b['value@rvs@secondary@fluxrv'], atol=0.3)
```
kecnry, line 28: these will run faster if you're explicit and always provide the tags (`set_value(qualifier='teff', value=10000)` and avoiding twigs).

phoebe/frontend/bundle.py
```diff
@@ -10383,6 +10388,24 @@ def restore_conf():
...
+        subset = []
+
+        if l3_needed:
+            subset += [ds for ds in datasets if len(self.filter('l3_mode', dataset=ds, check_visible=True)) > 0]
+
+        if mesh_needed:
+            subset += [ds for ds, kind in zip(datasets, ds_kinds) 
+                if kind == 'lc'
+                or kind == 'lp'
+                or (kind == 'rv' and len(self.filter(qualifier='rv_method', dataset=ds, compute=compute, value='flux-weighted', **_skip_filter_checks)) > 0)
```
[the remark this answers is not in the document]
aprsa: I don't think that would solve the issue; we need the opposite: even if value='dynamical', meshes are needed when kind != 'phoebe'. But certainly something to that extent, yes.

phoebe/frontend/bundle.py
```diff
@@ -10383,6 +10388,27 @@ def restore_conf():
 
         return system
 
+    def _datasets_where(self, compute, mesh_needed=False, l3_needed=False):
+        datasets = self.filter(compute=compute, qualifier='enabled', value=True, **_skip_filter_checks).datasets
+        ds_kinds = [self.filter(dataset=ds, context='dataset', **_skip_filter_checks).kind for ds in datasets]
+        backend = self.filter(compute=compute).kind
```
kecnry, line 10394:
```suggestion
        backend = self.filter(compute=compute, context='compute', **_skip_filter_checks).kind
```

phoebe/frontend/bundle.py
```diff
@@ -10440,6 +10460,12 @@ def compute_l3s(self, compute=None, use_pbfluxes={},
         if not isinstance(compute, str):
             raise TypeError("compute must be a single value (string)")
 
+        # either take user-passed datasets or datasets that have an l3_mode:
+        datasets = kwargs.pop('dataset', self._datasets_where(compute=compute, l3_needed=True))
```
kecnry, line 10464:
```suggestion
        datasets = kwargs.pop('dataset') if 'dataset' in kwargs else self._datasets_where(compute=compute, l3_needed=True)
```

phoebe/frontend/bundle.py
```diff
@@ -10457,10 +10483,9 @@ def compute_l3s(self, compute=None, use_pbfluxes={},
                                              **kwargs)
 
             # don't allow things like model='mymodel', etc
-            if not kwargs.get('skip_checks', False):
-                forbidden_keys = parameters._meta_fields_filter
-                compute_ps = self.get_compute(compute, **_skip_filter_checks)
-                self._kwargs_checks(kwargs, additional_allowed_keys=['system', 'skip_checks', 'ret_structured_dicts', 'pblum_method']+compute_ps.qualifiers, additional_forbidden_keys=forbidden_keys)
+            forbidden_keys = parameters._meta_fields_filter
+            compute_ps = self.get_compute(compute, **_skip_filter_checks)
+            self._kwargs_checks(kwargs, additional_allowed_keys=['system', 'skip_checks', 'ret_structured_dicts', 'pblum_method']+compute_ps.qualifiers, additional_forbidden_keys=forbidden_keys)
```
kecnry, line 10488: why the change here to not allowing skipping the checks?

phoebe/frontend/bundle.py
```diff
@@ -10383,6 +10388,27 @@ def restore_conf():
 
         return system
 
+    def _datasets_where(self, compute, mesh_needed=False, l3_needed=False):
+        datasets = self.filter(compute=compute, qualifier='enabled', value=True, **_skip_filter_checks).datasets
```
kecnry, line 10392:
```suggestion
        datasets = self.filter(compute=compute, context='compute', qualifier='enabled', value=True, **_skip_filter_checks).datasets
```

phoebe/frontend/bundle.py
```diff
@@ -10383,6 +10388,27 @@ def restore_conf():
...
+        ds_kinds = [self.filter(dataset=ds, context='dataset', **_skip_filter_checks).kind for ds in datasets]
+        backend = self.filter(compute=compute).kind
+
+        subset = []
+
+        if l3_needed:
+            subset += [ds for ds in datasets if len(self.filter('l3_mode', dataset=ds, check_visible=True)) > 0]
```
kecnry, line 10399:
```suggestion
            subset += [ds for ds in datasets if len(self.filter(qualifier='l3_mode', dataset=ds, context='dataset', check_visible=True)) > 0]
```

kecnry (commented):
I like this solution, just a few small suggestions to prevent potential issues when models/solutions exist and to help optimize some calls by being explicit in filter calls, then I think this is ready to go!  Thanks!

phoebe/frontend/bundle.py
aprsa: the previous `elif` already tested for `skip_checks`; see a couple of preceding lines of code.

phoebe/frontend/bundle.py
aprsa: Is this really needed?
```
>>> x = {'a': 1, 'b': 2}
>>> x.pop('c', 12)
12
>>> x
{'a': 1, 'b': 2}
>>> x.pop('a', 3)
1
>>> x
{'b': 2}
```

tests/tests/test_rvs/test_dynrvs.py
aprsa: certainly, but this is on the user-end and this code seems more readable to me, no? Either way, I have no strong feelings either way.

kecnry: We can always clean it up later - but do think it makes sense to try to make tests run efficiently so that CI doesn't take long (but we're talking 0.1 seconds here or something like that, so not a huge deal).

kecnry (approved):

PR #826, opened 2024-02-18
Interpolation/extrapolation switch to a C module

aprsa:
This is a complete rewrite of the ndpolator in C. There are a few more things I'd like to optimize but the code is fully functional and it passes all tests. As the first step, I'd kindly ask for the review of dependencies/ndpolator code. It is C but it is heavily documented. You can generate fully documentation using doxygen. @Raindogjones, the treatment of down-dimensioning has been reworked, so please see if this fixes the issue you were seeing before. I am leaving this in draft stage for now as we will likely need a few iterations to get right. This is a HUGE diff so brace yourselves...


aprsa:
Ignore test failures, they're all benign: we need to rebase to 2.4.13 once out (which will fix 3.7 failures) and the mac compiler is finicky about defining local variables, which is an easy fix.

kecnry:
running for the default binary, I'm not seeing:

```
File ~/repos/phoebe2/phoebe/frontend/bundle.py:11929, in Bundle.run_compute(self, compute, model, solver, detach, dataset, times, return_changes, **kwargs)
  11927 if 'lc' in ds_kinds_enabled or 'rv' in ds_kinds_enabled or 'lp' in ds_kinds_enabled:
  11928     logger.info("run_compute: computing necessary ld_coeffs, pblums, l3s")
> 11929     self.compute_ld_coeffs(compute=compute, skip_checks=True, set_value=True, **{k:v for k,v in kwargs.items() if k in computeparams.qualifiers})
  11930     # NOTE that if pblum_method != 'phoebe', then system will be None
  11931     # otherwise the system will be create which we can pass on to the backend
  11932     # the phoebe backend can then skip initializing the system at least on the master proc
  11933     # (workers will need to recreate the mesh)
  11934     system, pblums_abs, pblums_scale, pblums_rel, pbfluxes = self.compute_pblums(compute=compute, ret_structured_dicts=True, skip_checks=True, **{k:v for k,v in kwargs.items() if k in computeparams.qualifiers})

File ~/repos/phoebe2/phoebe/frontend/bundle.py:10328, in Bundle.compute_ld_coeffs(self, compute, set_value, **kwargs)
  10324 query_pts = np.array([[teff, logg, abun],])
  10326 # interpolate_ldcoeffs() always returns an array, so we need
  10327 # the first element of the array.
> 10328 ld_coeffs = pb.interpolate_ldcoeffs(
  10329     query_pts=query_pts,
  10330     ldatm=ldcs,
  10331     ld_func=ld_func,
  10332     intens_weighting=intens_weighting,
  10333     ld_extrapolation_method=ld_extrapolation_method
  10334 )[0]
  10336 # NOTE: these may return nans... if so, run_checks will handle the error
  10338 logger.info("interpolated {} ld_coeffs{}={}".format(ld_func, bol_suffix, ld_coeffs))

File ~/repos/phoebe2/phoebe/atmospheres/passbands.py:1385, in Passband.interpolate_ldcoeffs(self, query_pts, ldatm, ld_func, intens_weighting, ld_extrapolation_method)
   1382 if f'{ldatm}:ld' not in self.content:
   1383     raise ValueError(f'Limb darkening coefficients for ldatm={ldatm} are not available; please compute them first.')
-> 1385 ld_coeffs = self.ndp[ldatm].ndpolate(f'ld@{intens_weighting}', query_pts, extrapolation_method=ld_extrapolation_method)
   1386 return ld_coeffs[s[ld_func]]

File ~/repos/phoebe2/phoebe/dependencies/ndpolator/ndpolator.py:94, in Cndpolator.ndpolate(self, table, query_pts, extrapolation_method)
     92 capsule = self.table[table][2]
     93 if capsule:
---> 94     interps = cndpolator.ndpolate(capsule=capsule, query_pts=query_pts, nbasic=len(self.axes), extrapolation_method=extrapolation_method)
     95 else:
     96     attached_axes = self.table[table][0]

TypeError: function missing required argument 'axes' (pos 2)
```

aprsa:
Can someone confirm this? I cannot reproduce on my end, and all checks are passing on CI as well. @kecnry, could it be that cndpolator wasn't rebuilt properly?

phoebe/atmospheres/passbands.py
```diff
@@ -295,8 +295,8 @@ def __init__(self, ptf=None, pbset='Johnson', pbname='V', effwl=5500.0,
         # self.nntree = dict()              # nearest neighbor tree for blending and extrapolation
         # self.indices = dict()             # nearest neighbor indices for blending and extrapolation
         # self.ics = dict()                 # inferior corners for blending and extrapolation
-        self.blending_region = dict()     # blending regions
-        self.mapper = dict()              # mapping from model atmosphere axes to interpolation axes
+        # self.blending_region = dict()     # blending regions
```
Raindogjones, line 298: I'm guessing this is a temporary change to disable blending for now?

phoebe/atmospheres/passbands.py
```diff
@@ -631,48 +624,45 @@ def load(cls, archive, load_content=True, init_extrapolation=True):
...
-                            # Rebuild the table of non-null indices for the nearest neighbor lookup:
-                            # self.nntree[atm], self.indices[atm] = ndpolator.kdtree(self.atm_axes[atm][:-1], self.atm_photon_grid[atm][...,-1,:])
+                        # specific passband intensities:
+                        mus = np.array(list(hdul[f'{prefix}_mus'].data['mu']))
+                        self.ndp[atm].register('imu@photon', (mus,), self.atm_photon_grid[atm])
+                        self.ndp[atm].register('imu@energy', (mus,), self.atm_energy_grid[atm])
 
-                            # Rebuild blending map:
+                        if init_extrapolation:
+                            # CHECKME: probably obsolete
```
Raindogjones, line 649: ?

phoebe/atmospheres/passbands.py
```diff
@@ -631,48 +624,45 @@ def load(cls, archive, load_content=True, init_extrapolation=True):
                 for atm in ['ck2004', 'phoenix', 'tmap']:
```
Raindogjones, line 624: I'm wondering if we could do something similar for pb.save? This updated way of loading the passband is much more efficient when it comes to adding new atmospheres.  Can we update save and then make the list of atmospheres a global list that we define elsewhere? Like _supported_atms but without blackbody and extern_*?

phoebe/atmospheres/passbands.py
```diff
@@ -3084,22 +2986,22 @@ def Inorm_bol_bb(Teff=5772., logg=4.43, abun=0.0, atm='blackbody', intens_weight
     )
 
     pb.compute_blackbody_intensities(include_extinction=True)
-#    pb.compute_bb_reddening(verbose=True)
+    # pb.compute_bb_reddening(verbose=True)
```
Raindogjones, line 2989: I think we should just go ahead and delete these references to the old extinction calculations now these functions don't exist.

Raindogjones (commented):
I've reviewed passbands.py and made a couple of comments. In the mean time, I'll keep testing.

Raindogjones:
I just pushed a couple of small updates so that the required changes for the new TMAP grids are already there (more to make my life easier with the testing than anything else). Feel free to reverse the commits if you think they should be kept separate, but they don't change much other than splitting TMAP into four grids.

phoebe/atmospheres/passbands.py
aprsa: nope, that's a permanent change; blending regions are now replaced by unit hypercubes in cndpolator. My upcoming commit will remove these altogether.

phoebe/atmospheres/passbands.py
aprsa: Yes, that would be awesome! In fact, I want to move away from individual axes and tables and store everything in Ndpolator class instances. TBD.

phoebe/atmospheres/passbands.py
aprsa: Agreed, done.

phoebe/atmospheres/passbands.py
```diff
@@ -375,7 +369,7 @@ def save(self, archive, overwrite=True, update_timestamp=True, export_inorm_tabl
         header['TIMESTMP'] = timestamp
         header['PBSET'] = self.pbset
         header['PBNAME'] = self.pbname
-        header['EFFWL'] = self.effwl
+        header['EFFWL'] = self.effwl.value
```
kecnry, line 372: is this already guaranteed to be in meters?  Or should we do `to_value` to double check (I see later that meters are assumed).

phoebe/atmospheres/passbands.py
```diff
@@ -1142,6 +1121,7 @@ def compute_intensities(self, atm, path, impute=False, include_extinction=False,
         # self._ck2004_boosting_photon_grid = np.nan*np.ones((len(self.atm_axes['ck2004'][0]), len(self.atm_axes['ck2004'][1]), len(self.atm_axes['ck2004'][2]), len(self.atm_axes['ck2004'][3]), 1))
 
         if include_extinction:
+            # ? Should this not include mus as well?
```
kecnry, line 1124: should it?

phoebe/frontend/bundle.py
```diff
@@ -3991,7 +3991,7 @@ def ld_coeffs_len(ld_func, ld_coeffs):
                                             True, 'run_compute')
                         else:
                             atm = self.get_value(qualifier='atm', component=component, compute=compute, context='compute', atm=kwargs.get('atm', None), **_skip_filter_checks)
-                            if atm not in ['ck2004', 'phoenix', 'tmap']:
+                            if atm not in ['ck2004', 'phoenix', 'tmap_sdO', 'tmap_DA', 'tmap_DAO', 'tmap_DO']:
```
kecnry, line 3994: can we store this list somewhere in `passbands.py` and import it here?

kecnry (commented):
Just a few comments from skimming the code - I'll go back and test functionality more thoroughly in the next week or so and then let's get this in! 🤞

aprsa:
@kecnry, @Raindogjones, as you might have noticed, and in line of what was discussed at the telecon yesterday, I separated ndpolator from phoebe. The latest commit passes all the tests, so I think that the only remaining issue is the fits keyword length. Once that is done, I'll convert this draft PR to an actual PR and we can merge it in.

aprsa:
> TA = TMAP DA
> TO = TMAP DO
> TS = TMAP sdO
> TM = TMAP DAO

@Raindogjones, works for me!

phoebe/atmospheres/passbands.py
kecnry: as long as we aren't worried about ever deviating from that, then this is probably slightly cheaper.

phoebe/frontend/bundle.py
kecnry: if the list appears anywhere else, then I think consolidating to one place would be useful to avoid them falling out of sync, otherwise this is fine for now

Raindogjones: This is the same as _supported_atms in passbands.py just with blackbody and extern_* removed.  It is also defined multiple times in passbands.py when checking for how to handle limb-darkening etc., so I think we could consolidate that everywhere and just have it defined once in passbands.py

kecnry: ok, if that doesn't fit here @aprsa - please just create a follow-up ticket so we don't forget to address it eventually

aprsa: I added it to the project, but I don't think we should roll this into this PR.

phoebe/atmospheres/passbands.py
aprsa: I think it should, but this can also be deferred. Here's the [action item](https://github.com/orgs/phoebe-project/projects/2/views/11?pane=issue&itemId=57136993).

phoebe/frontend/bundle.py
aprsa: Done; action item is [here](https://github.com/orgs/phoebe-project/projects/2/views/11?pane=issue&itemId=57136481).

Raindogjones:
Requested TMAP changes made (plus references).  @aprsa will need to rebuild the default bundles and passbands but then I think that is it!

kecnry:
Once the shipped passbands are updated the errors should go away, but I'm curious if our warnings to update are sufficient or if we're now going to have to _require_ updating all passbands when updating phoebe.  Is there anyway we can keep this backwards compatible?

aprsa:
> Once the shipped passbands are updated the errors should go away, but I'm curious if our warnings to update are sufficient or if we're now going to have to _require_ updating all passbands when updating phoebe. Is there anyway we can keep this backwards compatible?

We won't need to require the update. This is only because the stock Johnson V passband didn't have blackbody extinction incorporated before (which in and of itself was an oversight). I'll edit the test that's failing and account for this omission, but otherwise there will be no adverse effects. We remain backwards-compatible.

phoebe/atmospheres/passbands.py
Raindogjones: Originally we decided to just derive the extinction corrections based on Inorms so as to not massively increased the passband file sizes.  Given the magnitude of the correction from extinction, I still think this is fine (and is the way we say that we do it in the paper).  Limb-darkening changes the SED but also reduces the brightness, so the majority of the flux comes from the centre where the SED is very, very similar to Inorm. Changing to Imu would would increase passband sizes by at least an order of magnitude for negligible gain.

kecnry (approved):

PR #832, opened 2024-03-12
Fix treatment of distance for alternate backends

### LINKED ISSUE #830: Treatment of distance using ellc
Hi everybody,
I was trying to do something akin to SED using ellc backend in Phoebe 2.4.11
It seems that fluxes scale as 1/d^4 instead of 1/d^2.
I tried to do as an example sun with a small companion
and test for a value of solar constant - code below.
To obtain reasonable flux around 1360 W/m^2 I came up with
b['distance@system']=np.sqrt(1.5e11/np.sqrt(3.14))
and that seems to work.
Do you have an idea what is going on?

Using phoebe backend gives correct values.

Thanks a lot,
Jozef Liptak

=====================================================================
```
import phoebe
from phoebe import u # units
import numpy as np

logger = phoebe.logger()

b= phoebe.default_binary()
b.set_value('teff@primary', 5770)
b.set_value('requiv@primary', 1.0)
b.set_value('period@binary', 365)
b.set_value('teff@secondary', 3500)
b.set_value('q', 0.08)
b.set_value('requiv@secondary', 0.1)
b.set_value('sma@binary', 215)
b.set_value('incl@binary', 90)

b.add_dataset('lc', compute_times=phoebe.linspace(0,1000,10), overwrite=True, dataset='lc_K')
b['passband@lc_K']='Bolometric:900-40000'
b['pblum_mode@lc_K@dataset']='absolute'
b['distance@system']=np.sqrt(1.5e11/np.sqrt(3.14))

b.add_compute('ellc', compute='fastcompute', overwrite=True)
b.set_value_all(qualifier='ld_mode', value='lookup')
b.set_value_all('ld_coeffs_source', value='phoenix')
b.set_value_all('atm', 'phoenix')
b.run_compute(model='init0', compute='fastcompute', overwrite=True)

afig, mplfig  = b.plot( x='phases', model='init0', save='test.pdf', show=True)
```

kecnry:
This PR avoids the duplicated handling of distance for alternate backends (except for legacy) that was erroneously introduced by #606.

This fixes #830

### DIFF
--- a/phoebe/frontend/bundle.py
+++ b/phoebe/frontend/bundle.py
@@ -12124,20 +12124,22 @@ class Bundle(ParameterSet):
                                 logger.debug("applying scale_factor={} to {} parameter in mesh".format(scale_factor, mesh_param.qualifier))
                                 mesh_param.set_value(mesh_param.get_value()*scale_factor, ignore_readonly=True)
 
-                # handle flux scaling based on distance and l3
-                # NOTE: this must happen AFTER dataset scaling
-                distance = self.get_value(qualifier='distance', context='system', unit=u.m, **_skip_filter_checks)
-                for flux_param in ml_params.filter(qualifier='fluxes', kind='lc', **_skip_filter_checks).to_list():
-                    dataset = flux_param.dataset
-                    if dataset in datasets_dsscaled:
-                        # then we already handle the scaling (including l3)
-                        # above in dataset-scaling
-                        continue
+                # alternate backends other than legacy already account for distance via pbflux
+                if computeparams.kind in ['phoebe', 'legacy']:
+                    # handle flux scaling based on distance and l3
+                    # NOTE: this must happen AFTER dataset scaling
+                    distance = self.get_value(qualifier='distance', context='system', unit=u.m, **_skip_filter_checks)
+                    for flux_param in ml_params.filter(qualifier='fluxes', kind='lc', **_skip_filter_checks).to_list():
+                        dataset = flux_param.dataset
+                        if dataset in datasets_dsscaled:
+                            # then we already handle the scaling (including l3)
+                            # above in dataset-scaling
+                            continue
 
-                    fluxes = flux_param.get_value(unit=u.W/u.m**2)
-                    fluxes = fluxes/distance**2 + l3s.get(dataset)
+                        fluxes = flux_param.get_value(unit=u.W/u.m**2)
+                        fluxes = fluxes/distance**2 + l3s.get(dataset)
 
-                    flux_param.set_value(fluxes, ignore_readonly=True)
+                        flux_param.set_value(fluxes, ignore_readonly=True)
 
                 # handle vgamma and rv_offset
                 vgamma = self.get_value(qualifier='vgamma', context='system', unit=u.km/u.s, **_skip_filter_checks)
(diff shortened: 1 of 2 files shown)


kecnry:
Seems the tests are failing when attempting to generate a synthetic light curve with jktebop - will have to investigate further or fallback to just testing ellc for now.

phoebe/frontend/bundle.py
```diff
@@ -12124,20 +12124,22 @@ def _scale_fluxes_cfit(fluxes, scale_factor):
...
-                # NOTE: this must happen AFTER dataset scaling
-                distance = self.get_value(qualifier='distance', context='system', unit=u.m, **_skip_filter_checks)
-                for flux_param in ml_params.filter(qualifier='fluxes', kind='lc', **_skip_filter_checks).to_list():
-                    dataset = flux_param.dataset
-                    if dataset in datasets_dsscaled:
-                        # then we already handle the scaling (including l3)
-                        # above in dataset-scaling
-                        continue
+                # alternate backends other than legacy already account for distance via pbflux
+                if computeparams.kind in ['phoebe', 'legacy']:
```
aprsa, line 12128: so we don't need to do anything if the backend isn't one of these?

tests/tests/test_distance/test_distance.py
```diff
@@ -0,0 +1,39 @@
+"""
+"""
+import phoebe
+
+
+def test_distance_scaling_alt_backend(verbose=False, plot=False):
+    b = phoebe.Bundle.default_binary()
```
aprsa, line 7: is there a reason you prefer a call to Bundle even though default_binary() is exposed in top-level?

tests/tests/test_distance/test_distance.py
```diff
@@ -0,0 +1,39 @@
...
+
+    b.add_dataset('lc', compute_phases=[0.25])
+    for backend in phoebe.list_available_computes():
+        if backend == 'phoebe':
+            # compute options already exist
+            continue
+        if backend == 'jktebop':
+            # TODO: re-enable this in the test, currently fails on CI
```
aprsa, line 15: does this limit the use of jktebop for distances?

aprsa (approved):

tests/tests/test_distance/test_distance.py
kecnry: no, I just copy-pasted from another test that was probably written before the shortcut.  They're identical though.

tests/tests/test_distance/test_distance.py
kecnry: no, it just means we don't have test coverage.  The problem is with _running_ jktebop on CI, not necessarily the distance treatment.

phoebe/frontend/bundle.py
kecnry: phoebe and legacy were not affected by the bug, if that's what you're asking.  All other alternate backends which use pbflux were double-accounting for distance, but do not any longer.

tests/tests/test_distance/test_distance.py
aprsa: it would be nice to have a consistent treatment throughout the test suite, so my suggestion would be to decide and stick with one, especially in the newly written code.

aprsa: I've done most of that work already when I ensured pytest compatibility, so I don't think it'll be nearly as daunting to finish. But at least for newly added tests let's try to be consistent?

PR #837, opened 2024-03-28
run_checks_compute() bugfix

aprsa:
The run_checks_compute() function included the internal "_default" compute parameter set in checking the ck2004 model atmosphere table existence in the passband files. This caused the checks to fail even if ck2004 was never needed.

### DIFF
--- a/phoebe/frontend/bundle.py
+++ b/phoebe/frontend/bundle.py
@@ -3754,7 +3754,6 @@ class Bundle(ParameterSet):
         all_pbs = list_passbands(full_dict=True)
         online_pbs = list_online_passbands(full_dict=True)
 
-        # TODO: is this robust against flipping constraints?
         pb_needs_ext = self.get_value(qualifier='ebv', context='system', **_skip_filter_checks) != 0
 
         for pbparam in self.filter(qualifier='passband', **_skip_filter_checks).to_list():
@@ -3772,7 +3771,6 @@ class Bundle(ParameterSet):
 
             missing_pb_content = []
 
-            # TODO: do we really want to run this every time we run checks? Perhaps move to init?
             if pb_needs_ext and pb in ['Stromgren:u', 'Johnson:U', 'SDSS:u', 'SDSS:uprime']:
                 # need to check for bugfix in coefficients from 2.3.4 release
                 installed_timestamp = installed_pbs.get(pb, {}).get('timestamp', None)
@@ -3783,10 +3781,10 @@ class Bundle(ParameterSet):
                                     True, 'run_compute')
 
             # NOTE: atms are not attached to datasets, but per-compute and per-component
-            for atmparam in self.filter(qualifier='atm', kind='phoebe', compute=computes, **_skip_filter_checks).to_list() + self.filter(qualifier='ld_coeffs_source').to_list():
-                if '_default' in atmparam.twig:
-                    continue
-
+            # NOTE: atmparam includes a '_default' compute pset, which depends on the
+            # ck2004 atmospheres; the checks should not include it. This is achieved
+            # by filtering check_visible=True in the line below:
+            for atmparam in self.filter(qualifier='atm', kind='phoebe', compute=computes, check_visible=True, check_default=False).to_list() + self.filter(qualifier='ld_coeffs_source').to_list():
                 # check to make sure passband supports the selected atm
                 atm = atmparam.get_value(**_skip_filter_checks)
                 if atmparam.qualifier == 'ld_coeffs_source' and atm == 'auto':
--- a/phoebe/frontend/bundle.py
+++ b/phoebe/frontend/bundle.py
@@ -1,7 +1,6 @@
 import sys
 import subprocess
 import os
-import numpy as np
 
 try:
     from subprocess import DEVNULL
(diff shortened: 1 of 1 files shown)


aprsa:
Note that this fix had to be applied to the blending-dev branch so that other commits can be tested/applied.

phoebe/frontend/bundle.py
```diff
@@ -3765,11 +3767,12 @@ def run_checks_compute(self, compute=None, solver=None, solution=None, figure=No
 
             pb_needs_Inorm = True
             pb_needs_Imu = True
-            pb_needs_ld = True #np.any([p.get_value()!='interp' for p in self.filter(qualifier='ld_mode', dataset=pbparam.dataset, context='dataset', **_skip_filter_checks).to_list()])
+            pb_needs_ld = True
             pb_needs_ldint = True
 
             missing_pb_content = []
 
+            # TODO: do we really want to run this every time we run checks? Perhaps move to init?
```
kecnry, line 3775: yes, because the selected passband is allowed to change.  We could probably do some smarter caching and only checking changed parameters, etc, but that would require quite a bit of refactoring.

phoebe/frontend/bundle.py
```diff
@@ -3753,6 +3754,7 @@ def run_checks_compute(self, compute=None, solver=None, solution=None, figure=No
         all_pbs = list_passbands(full_dict=True)
         online_pbs = list_online_passbands(full_dict=True)
 
+        # TODO: is this robust against flipping constraints?
```
kecnry, line 3757: it should be, yes

phoebe/frontend/bundle.py
```diff
@@ -1,6 +1,7 @@
 import sys
 import subprocess
 import os
+import numpy as np
```
kecnry, line 4: this seems to be unused in this PR

phoebe/frontend/bundle.py
```diff
@@ -3781,6 +3784,8 @@ def run_checks_compute(self, compute=None, solver=None, solution=None, figure=No
 
             # NOTE: atms are not attached to datasets, but per-compute and per-component
             for atmparam in self.filter(qualifier='atm', kind='phoebe', compute=computes, **_skip_filter_checks).to_list() + self.filter(qualifier='ld_coeffs_source').to_list():
+                if '_default' in atmparam.twig:
+                    continue
```
kecnry, line 3788:
```suggestion
            for atmparam in self.filter(qualifier='atm', kind='phoebe', compute=computes, check_visible=True, check_default=False).to_list() + self.filter(qualifier='ld_coeffs_source').to_list():
```

kecnry (commented):
Any place we can add a quick regression test?

phoebe/frontend/bundle.py
aprsa: I adopted this suggestion but I find it *very* obfuscated (not in a million years would I recognize that `check_visible=True` skips the '_default' pset). So, as a compromise, I added a comment in the code that explains this.

aprsa:
> Any place we can add a quick regression test?

It's implicitly tested in the blackbody extinction test, so we do have coverage.

phoebe/frontend/bundle.py
```diff
@@ -3780,8 +3781,10 @@ def run_checks_compute(self, compute=None, solver=None, solution=None, figure=No
                                     True, 'run_compute')
 
             # NOTE: atms are not attached to datasets, but per-compute and per-component
-            for atmparam in self.filter(qualifier='atm', kind='phoebe', compute=computes, **_skip_filter_checks).to_list() + self.filter(qualifier='ld_coeffs_source').to_list():
-
+            # NOTE: atmparam includes a '_default' compute pset, which depends on the
+            # ck2004 atmospheres; the checks should not include it. This is achieved
+            # by filtering check_visible=True in the line below:
+            for atmparam in self.filter(qualifier='atm', kind='phoebe', compute=computes, check_visible=True, check_default=False).to_list() + self.filter(qualifier='ld_coeffs_source').to_list():
```
kecnry, line 3787: woops, I actually think I got these Trues and Falses backwards.  We want to skip the visible checks but include the default checks so that `_default` are excluded from the returned parameters entirely (didn't notice it until I saw your comment explanation! 😬 )

phoebe/frontend/bundle.py
kecnry: we get it from importing all from parameters, right?  (Not really a circular import... but importing `*` surely isn't the best, let's make a ticket for that to refactor and explicitly import things, but that definitely shouldn't happen here)

kecnry:
> It's implicitly tested in the blackbody extinction test in PR https://github.com/phoebe-project/phoebe2/pull/826, so we do have coverage.

We'll have coverage whenever that is released... but I guess its unlikely this will be reintroduced, especially with the comment explaining the reason for using the default checks.

phoebe/frontend/bundle.py
aprsa: ok, so do we want to do it this way then? As this way buries the logic deeper, it might be difficult to parse. Even now I'm not sure: if check_visible=True, I'd interpret that to only take visible psets, which would exclude _default. Now you're saying that check_visible should be False? And then that check_default should be True? That seems counter-intuitive to me.

kecnry: Yes, it will be (marginally) cheaper, so I think since this runs often it is worth the explanatory comment and keeping the code as efficient as possible.

aprsa: the problem is that even the explanatory comment doesn't help me understand the logic. ;) But okay, I'll recommit with the fix and move on.

kecnry: `check_visible` checks whether a parameter is visible, and if it is not, excludes it from the results.  `check_default` similarly checks whether a parameter has any tag (compute, component, dataset, etc) as `_default`, and if so, excludes it from the results.  In this case you do NOT want parameters that are tagged default, so we need to use `check_default=True` (or skip the parameters with a continue).  We are not worried about parameters that are not visible/relevant, so we can save some time by passing `check_visible=False`.

kecnry (approved):
Thanks! Just for peace of mind - can you please copy this finalized version over to the other branch and confirm that it passes the test coverage there before we merge?

aprsa:
> Thanks! Just for peace of mind - can you please copy this finalized version over to the other branch and confirm that it passes the test coverage there before we merge?

Yep, it passes. Merge away.

PR #844, opened 2024-04-10
Fixes a cryptic error message when query points are out of bounds.

### LINKED ISSUE #843: Cryptic error messages for atm out-of-bounds
As part of the passbands/ndpolator rewrite, the error messages have become more cryptic and less useful.
For example,
`import phoebe`
`b=phoebe.default_binary()`
`b['teff@primary']=50000`
`b.add_dataset('lc',time=0)`
`b.run_compute()`
results in  `ValueError: irradiation resulted in nans for teffs` rather than raising an out-of-bounds error for the primary atmosphere.

Turning off irradiation with `b.run_compute(irrad_method='none')` changes the error message but still not in a useful way. Either raising no error and populating the light curve with NaNs or throwing `ValueError: array must not contain infs or NaNs` if pblum_mode is, e.g., dataset-scaled

aprsa:
Closes #843.

### DIFF
--- a/phoebe/atmospheres/passbands.py
+++ b/phoebe/atmospheres/passbands.py
@@ -1620,7 +1620,9 @@ class Passband:
         """
 
         log10_Inorm = self.ndp[atm].ndpolate(f'inorm@{intens_weighting}', query_pts, extrapolation_method=atm_extrapolation_method)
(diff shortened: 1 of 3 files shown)


kecnry (approved):
works for me!

PR #847, opened 2024-04-17
backwards compatibility tests for phoebe 2.4 -> 2.5

kecnry:
this adds compatibility tests for importing phoebe bundles created in phoebe 2.4.x into phoebe 2.5.


tests/tests/test_backward_compatibility/test_import.py
```diff
@@ -131,7 +164,11 @@ def test_23(verbose=False, plot=False):
         if parse(phoebe.__version__) >= parse("2.3.0") and parse(phoebe.__version__) < parse("2.4.0"):
             _export_23('23_export.phoebe')
             exit()
+        if parse(phoebe.__version__) >= parse("2.4.0") and parse(phoebe.__version__) < parse("2.5.0"):
+            _export_24('24_export.phoebe')
+            exit()
```
aprsa, line 169: do we need all exit()s here?

aprsa (approved):

tests/tests/test_backward_compatibility/test_import.py
kecnry: no, but its convenient to just flip the one switch to call the script to create the bundles and skip the tests themselves 🤷. Alternatively I guess the tests could be in an else

Issue #850, opened 2024-04-20
libphoebe function roche_misaligned_marching_mesh segmentation fault on MacOS


bpablo:
The following code fails on Mac with two different machines but works on windows:

```
import numpy as np
import libphoebe
values = np.array([3.13850013e-18, 9.30808211e-33, 1.00000000e+00])

libphoebe.roche_misaligned_marching_mesh(0.6, 1.0, 0.8140441018004452, values, 6.056472707118865, 0.026244734687742324)
```

horvatm:
Hi Bert,

long time no see. Which phoebe version are you using? Is this a specific Apple architecture, Mx models perhaps? With what errors does it fail? Can you increase verbosity
~~~
libphoebe.setup_verbosity(4)
~~~
and run it again? Can I get an account on such a machine, so that I can see where it fails?

On my Ubuntu 22.04,  Intel i7 machine using current master branch it works without problems.

Cheers, Martin

Raindogjones:
Couldn't reproduce on my mac (2018 Intel running OS X 13.6.6 Ventura).

- [ ]

kecnry:
Neither can I on 12.6.5 Monterey with M1 Max in a conda env compiled with clang 14.0.6

bpablo:
Hey Martin,

I have an M1 Pro chip so arm architecture. I have a student using the same chip that can reproduce. I am running 14.4.1 and Xcode 15, which are new but I updated to see if I could fix the problem. Unsure exactly what I was running before, but there was no change. I tried pip installing locally and with conda as well as source installing locally and with conda, both of which give the same result.  Here is the full code with the verbosity flag:
import numpy as np
import libphoebe
libphoebe.setup_verbosity(4)
values = np.array([3.13850013e-18, 9.30808211e-33, 1.00000000e+00])
 libphoebe.roche_misaligned_marching_mesh(0.6, 1.0, 0.8140441018004452, values, 6.056472707118865, 0.026244734687742324)
roche_misaligned_marching_mesh::START
roche_misaligned_marching_mesh::spin:3.1385001300000000e-18 9.3080821100000003e-33 1.0000000000000000e+00 Omega:6.0564727071188651e+00 q=5.9999999999999998e-01 F=1.0000000000000000e+00 d=8.1404410180044517e-01 delta =2.6244734687742324e-02
 full=1 max_triangles=10000000
[...]

PR #854, opened 2024-04-25
Memory optimization

aprsa:
This PR eliminates duplication in memory consumption between the internal passband structures and ndpolator structures. It also optimizes passband loading/saving. Addresses issue [68](https://github.com/phoebe-project/phoebe-development/issues/68).


Raindogjones:
@aprsa Just tried to reproduce with python3.11 on my intel mac and no problems - suspect this might be related to #850 and specific to M1/ARM64 macs...

Raindogjones:
Paging @bpablo to see if it fails for him on his M1

bpablo:
I can reproduce. There is at least one test that seg faults: dperdt with the following error:
Segmentation fault: 11
/Users/bertpablo/.pyenv/versions/3.11.9/lib/python3.11/multiprocessing/resource_tracker.py:254: UserWarning: resource_tracker: There appear to be 1 leaked semaphore objects to clean up at shutdown
  warnings.warn('resource_tracker: There appear to be %d '

kecnry:
For this PR, I think we can safely ignore the failures as those are being handled in #850 and don't seem to have been made worse here (so let's keep any additional conversation in #850 and review this as-is).

phoebe/atmospheres/passbands.py
```diff
@@ -1631,46 +1363,46 @@ def _log10_Inorm(self, query_pts, atm, intens_weighting='photon', atm_extrapolat
...
-            log10_Inorm_bl = np.empty_like(teffs[nanmask])
-            for si, selem in enumerate(nv):
-                ic = [np.searchsorted(naxes[k], selem[k])-1 for k in range(len(naxes))]
-                seps = (np.abs(ndp.ics[atm]-np.array(ic))).sum(axis=1)
-                corners = np.argwhere(seps == seps.min()).flatten()
+            # log10_Inorm_bl = np.empty_like(teffs[nanmask])
+            # for si, selem in enumerate(nv):
+            #     ic = [np.searchsorted(naxes[k], selem[k])-1 for k in range(len(naxes))]
+            #     seps = (np.abs(ndp.ics[atm]-np.array(ic))).sum(axis=1)
+            #     corners = np.argwhere(seps == seps.min()).flatten()
```
kecnry, line 1373: (and all following sections) can we just delete these instead of commenting them out?  They'll always be in the git history if we need to find them to add them back in?

.gitignore
```diff
@@ -7,7 +7,7 @@
 /phoebe/atmospheres/tables/phoenix*
 /phoebe/atmospheres/tables/tmap*
 /phoebe/atmospheres/tables/tlusty
-/phoebe/atmospheres/tables/phoenix.orig
+/tests/tests/test_passbands/tables
```
kecnry, line 10: nothing seems to be written to this directory by the code, so I'm not sure it makes sense to add to the gitignore

phoebe/atmospheres/passbands.py
```diff
@@ -421,202 +414,45 @@ def save(self, archive, overwrite=True, update_timestamp=True, export_inorm_tabl
...
+                teffs, loggs, abuns, mus = self.ndp[atm].axes + self.ndp[atm].table['imu@photon'][0]
+                data.append(fits.table_to_hdu(Table({'teff': teffs}, meta={'extname': f'{prefix}_TEFFS'})))
+                data.append(fits.table_to_hdu(Table({'logg': loggs}, meta={'extname': f'{prefix}_LOGGS'})))
+                data.append(fits.table_to_hdu(Table({'abun': abuns}, meta={'extname': f'{prefix}_ABUNS'})))
+                data.append(fits.table_to_hdu(Table({'mu': mus}, meta={'extname': f'{prefix}_MUS'})))
+
+                if f'{atm}:ext' in self.content:
+                    ebvs, rvs = self.ndp[atm].table['ext@photon'][0]
+                    data.append(fits.table_to_hdu(Table({'ebv': ebvs}, meta={'extname': f'{prefix}_EBVS'})))
+                    data.append(fits.table_to_hdu(Table({'rv': rvs}, meta={'extname': f'{prefix}_RVS'})))
```
kecnry, line 429: yes!! 👏

phoebe/atmospheres/passbands.py
```diff
@@ -643,6 +479,15 @@ def load(cls, archive, load_content=True, init_extrapolation=True):
...
         logger.debug("loading passband from {}".format(archive))
 
+        atms = {
+            'ck2004': 'CK',
+            'phoenix': 'PH',
+            'tmap_sdO': 'TS',
+            'tmap_DA': 'TA',
+            'tmap_DAO': 'TM',
+            'tmap_DO': 'TO'
+        }
```
kecnry, line 489: this is in multiple places, can we put it somewhere common and import it?

tests/tests/test_passbands/test_passbands.py
```diff
@@ -0,0 +1,42 @@
+import phoebe
+from phoebe import u
+import os
+
+
+path = os.path.dirname(__file__)
+
+
+def test_compute(atm_grids_available=False, atm_path=None, atms=['ck2004', 'phoenix', 'tmap_sdO', 'tmap_DA', 'tmap_DAO', 'tmap_DO']):
```
kecnry, line 9: we could also import `atms` here to ensure any ones added in the future get coverage

tests/tests/test_passbands/test_passbands.py
```diff
@@ -0,0 +1,42 @@
...
+            pb.compute_intensities(atm=atm, path=f'{atm_path}/{atm}', verbose=True)
+            pb.compute_ldcoeffs(ldatm=atm)
+            pb.compute_ldints(ldatm=atm)
+
+    pb.save(os.path.join(path, 'test.fits'))
+
+
+def test_load():
+    pb = phoebe.atmospheres.passbands.Passband.load(os.path.join(path, 'test.fits'))
+    print(pb.content)
```
kecnry, line 33: (or put it behind some verbose flag that is off by default)

```suggestion
```

kecnry (commented):
this will be great to get in - thanks for the cleanup!

phoebe/atmospheres/passbands.py
aprsa: I am tackling this next and it's useful to have it in front of me rather than fishing through git history. I'll remove it in my next (related) PR.

.gitignore
aprsa: that's right, but this is a symbolic link to the (extensive) model atmosphere tables that are needed for the passband test. The logic there is to see whether tables are present and, if so, to run the test, otherwise skip it. Either way, git should not try to index those files, hence the entry in .gitignore.

phoebe/atmospheres/passbands.py
aprsa: yep, done; will be reflected in the next commit.

tests/tests/test_passbands/test_passbands.py
aprsa: changed to assertion to test whether 'blackbody' is in pb.content.

tests/tests/test_passbands/test_passbands.py
```diff
@@ -0,0 +1,43 @@
...
+    pb = phoebe.atmospheres.passbands.Passband.load(os.path.join(path, 'test.fits'))
+    assert 'blackbody:Inorm' in pb.content
+
+    # cleanup
+    os.remove(os.path.join(path, 'test.fits'))
+
+
+if __name__ == '__main__':
+    atm_path = os.path.join(path, 'tables')
+    test_compute(atm_grids_available=True, atm_path=atm_path)
```
kecnry, line 42: Maybe something like this so this doesn't fail on machines that don't have the directory?

```suggestion
    atm_path = os.path.join(path, 'tables')
    test_compute(atm_grids_available=os.path.isdir(atm_path), atm_path=atm_path)
```

PR #856, opened 2024-04-25
TST: segfault on macs

bpablo:
**NOT INTENDED TO MERGE**

This sets up a CI environment to try to reproduce the issues in #850, but disabling other tests so that it runs quicker (and therefore should not be merged).


kecnry:
Thanks for wrapping up the shell access @aprsa!  After we investigate this particular bug, we should consider merging this into a dedicated debug branch that we can just rebase onto main whenever needed!

PR #861, opened 2024-04-30
Fix CI on blending branch

kecnry:
This PR fixes issues with tests on the `feature-blending` branch.

Note: mac runners are still expected to fail here, but linux runners should be ✅ 🤞

### DIFF
--- a/tests/tests/test_passbands/test_passbands.py
+++ b/tests/tests/test_passbands/test_passbands.py
@@ -7,7 +7,7 @@ import os
 path = os.path.dirname(__file__)
 
 
-def test_compute(atm_grids_available=False, atm_path=None, atms=supported_atms.keys()):
+def test_compute(atm_grids_available=False, atm_path=None, atms=supported_atms):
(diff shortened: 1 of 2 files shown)


tests/tests/test_passbands/test_passbands.py
```diff
@@ -7,7 +7,7 @@
 path = os.path.dirname(__file__)
 
 
-def test_compute(atm_grids_available=False, atm_path=None, atms=supported_atms.keys()):
+def test_compute(atm_grids_available=False, atm_path=None, atms=supported_atms):
```
aprsa: ah, thanks for catching this; I believe it should be `atms=atm_tables.keys()` instead.

aprsa (approved):

tests/tests/test_passbands/test_passbands.py
aprsa: nope -- blackbody atms are already included (see the body of the function) and extern_* are only indexed, not computed. I'm actually thinking it might be time to retire extern_*.

PR #863, opened 2024-05-03
Passband cleanup with extinction bugfix

aprsa:
* when generating passband files, extinction tables were not stored correctly because of a missing "f" in the string definition
* passband history and comments were intertwined in a way that did not make sense. Now history is reserved for automated tasks and comments are reserved for user-provided information.
* minor code cleanup.

Related to [issue 73](https://github.com/phoebe-project/phoebe-development/issues/73) and [issue 65](https://github.com/phoebe-project/phoebe-development/issues/65).

### DIFF
--- a/phoebe/atmospheres/passbands.py
+++ b/phoebe/atmospheres/passbands.py
@@ -138,7 +138,7 @@ def raise_out_of_bounds(nanvals, atm=None, ldatm=None, intens_weighting=None):
 class Passband:
     def __init__(self, ptf=None, pbset='Johnson', pbname='V',
                  wlunits=u.AA, calibrated=False, reference='', version=1.0,
-                 comments='', oversampling=1, ptf_order=3, from_file=False):
+                 comment=None, oversampling=1, ptf_order=3, from_file=False):
         """
         <phoebe.atmospheres.passbands.Passband> class holds data and tools for
         passband-related computations, such as blackbody intensity, model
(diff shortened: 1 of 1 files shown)


kecnry:
Is any of this applicable for a bugfix release so that we can update passbands on the server with extinction compatible with 2.4?

aprsa:
Probably not worth the bugfix, no, because it only affects creating passbands, not hosting/parsing/loading them. I just created stock passbands (bolometric and V) and I'll soon create all others. They are backwards-compatible.

Raindogjones:
I don't think this bug affects the release version, so doesn't explain why extinction is missing from some passbands on the server.

aprsa:
@Raindogjones who says I was using the main branch when creating those passbands? ;)

Raindogjones:
Not me, that's for sure. Still means no bugfix necessary for 2.4

Raindogjones (approved):
All looks good to me.

phoebe/atmospheres/passbands.py
```diff
@@ -354,9 +373,9 @@ def save(self, archive, overwrite=True, update_timestamp=True, export_inorm_tabl
             have been deprecated since phoebe 2.4; for backwards compatibility
             we still may need to export them to fits files so that pre-2.4
             versions can use the same passband files.
-        * `history_entry` (string, optional): history entry to append to the
-            fits file.  Note that previous entries will be maintained if
-            (and only if) overwriting an existing file with `overwrite=True`.
+        * `export_legacy_comments` (bool, optional, default=True): whether to
+            export the dummy COMMENTS card (used in old passbands) in addition
+            to the new COMMENT card.
```
kecnry, line 378: is there a reason we can't just keep calling it `COMMENTS`?  Would that not be backwards-compatible either?

aprsa: yep, COMMENT is a reserved fits keyword that should be used exclusively because fits handles multiple instances of COMMENT natively under the hood. COMMENTS, on the other hand, is not standard and we need to implement the parsing logic in phoebe, which is a pain and it also violates fits standards. So I'd say we should stick to COMMENT. More on that [here](https://docs.astropy.org/en/stable/io/fits/index.html).

kecnry (approved):

PR #875, opened 2024-05-29
Bugfix - datasets now converted to list as intended in compute_ld_coeffs

MichaelAbdul-Masih:
In bundle.py the function `compute_ld_coeffs` was missing logic to convert datasets into a list before trying to append an additional dataset.  This caused an error pointed out in the following issue: https://github.com/phoebe-project/phoebe2-docs/issues/36

We have added the missing logic and now the error no longer occurs.


kecnry (approved):
Thanks for the quick fix!

PR #884, opened 2024-06-04
macos-14 segfault bugfix (marching)

aprsa:
This is @horvatm's proposed fix for meshing that caused segfaults on macos-14 (mac silicon, arm64).


aprsa:
**NOTE:** this PR needs to be cleaned up first to remove the debugging runners.

horvatm:
The code can be additionally cleaned, but I need to know that the current fix works with as many users as possible and passes all tests.

kecnry:
We'll need to revert all the changes to CI infrastructure before review/merge (although it would be nice to maintain the capability to enter a session in demand, probably as a separate PR or branch though).

kecnry:
This is now manually rebased on top of `bugfix-2.4.14` so that full CI can run.

setup.py
```diff
@@ -6,7 +6,7 @@
         'libphoebe',
         sources=['phoebe/lib/libphoebe.cpp'],
         language='c++',
-        extra_compile_args=["-std=c++11"],
+        extra_compile_args=["-std=c++14"],
```
kecnry, line 9: this will require an update to the install docs (and possibly the CI builds?).  Is this necessary for the fix or just to update to something more modern?

We currently say [on the install page](http://phoebe-project.org/install):

```
The C++ code in PHOEBE requires a compiler that supports C++11, either of the following should build correctly:

g++ (g++-5 or newer)
clang (clang-3.3 or newer)
```

What should this be updated to?

horvatm: The influence of this compiler flags is strange on macos. If you don't set it, macos starts to use C-compiler instead of C++-compiler. So it seems it needs a standard of C++ to be specified to even work properly. On a linux machine, it always uses the latest c++ standard, if c++ standad is not specified.  I suggest that we upgrade the standard to at least to c++14 so that macos is happy and linux also works.

bpablo: I also moved to the bugfix-2.4.14 branch after reading Kyle's previous comment and that works as well.

kecnry: I think this needs to bump the requirements in the installation page for clang to clang-3.4 (gcc-5 should still support c++14, as far as I can tell).  I'll make the necessary updates to the website.

PR #901, opened 2024-06-13
Updated functionality for lnf_rv

### LINKED ISSUE #752: sigmas_lnf for radial velocities with system velocity, and request for a simpler "jitter" term
Hello!

This is a question about the fractional increase of uncertainties through the "sigmas_lnf" parameter.

For radial velocity data, it seems to me that it scales based on the full radial velocity model; meaning that if the system radial velocity is not 0 km/s, the uncertainty will be applied asymmetrically to the RV curve.

Is that correct? I've looked at the code on https://github.com/phoebe-project/phoebe2/blob/7bc4489a5b2bd3f349291deec850589537109c0b/phoebe/parameters/parameters.py#LL3858C18-L3858C18 
I cannot find any clear indication that the system RV is "ignored" when performing this rescaling.

I think this should be looked into. If the system RV is not ignored, then I suspect the fractional uncertainty is not applied in a way that takes the Keplerian RV signal properly into account.

Separately, I have a feature request. Some exoplanet people forego the use of a fractional RV uncertainty in favor of a less conservative additive component, simply a jitter term that can be sampled over as a constant quadrature addition to all RV uncertainties:

`sigma**2 = sigma_n**2 + jitter**2`

Would you be interested in implementing that feature into PHOEBE?

kmhambleton:
Updated _calculate_cf function to adjust lnf calculation for radial velocities. The noise nuisance parameter for rvs no longer contains the rv amplitude.  

Closes #752 and #895

TODO:
- [ ] update documentation to mention change (and remove in release-2.5 branch)
- [x] add warning when using lnf with rv to mention change in behavior (and remove in release-2.5 branch)

### DIFF
--- a/phoebe/frontend/bundle.py
+++ b/phoebe/frontend/bundle.py
@@ -4615,6 +4615,13 @@ class Bundle(ParameterSet):
                                             +addl_parameters,
                                             True, 'run_solver')
 
+                # this check can/should be removed in PHOEBE 2.5
+                for param in self.filter(qualifier='sigmas_lnf', dataset=rv_datasets, context='dataset', **_skip_filter_checks).to_list():
+                    if np.isfinite(param.get_value()):
+                        report.add_item(self,
+                                        "behavior of sigmas_lnf for RVs was changed (fixed) in PHOEBE 2.4.15 to be independent of the RV value.  See https://github.com/phoebe-project/phoebe2/pull/901",
+                                        [param]+addl_parameters,
+                                        False, 'run_solver')
 
 
             if 'lc_datasets' in solver_ps.qualifiers:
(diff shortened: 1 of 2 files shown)


kecnry:
looks like the base bugfix branch (#907) is also failing CI, so likely due to upstream changes.  We can move forward with this if those errors are all the same, but then should probably try to get CI passing again before releasing.

kecnry:
Confirmed that the failing tests were the same as on the base branch.  @kmhambleton does the warning I added look reasonable to you (since this is a behavior change and we want to make sure anyone that was relying on the old behavior for some reason will see this change - we can remove the warning in the next minor release)?

kecnry (approved):
Going to assume the warning message is ok and merge this so that tests can run on the bugfix branch.

PR #914, opened 2024-06-27
Adding back the definition of self.wl to passband load() function

Raindogjones:
At some point (I haven't managed to track down when), the definition of self.wl disappeared from the passband load() function. This means that already created passbands cannot be built upon and needed to be initialised from scratch, i.e.
```python
pb=phoebe.get_passband("Johnson:V")
pb.compute_blackbody_response()
```
will fail, even though the passband includes all the required information to recalculate the blackbody response.

### DIFF
--- a/phoebe/atmospheres/passbands.py
+++ b/phoebe/atmospheres/passbands.py
@@ -433,6 +433,7 @@ class Passband:
             self.pbname = header['pbname']
             self.effwl = header['effwl']
             self.calibrated = header['calibrtd']
+            self.wl_oversampling = header.get('wlovsmpl', 1)
             self.comments = header['comments']
             self.reference = header['referenc']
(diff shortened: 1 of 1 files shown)


aprsa (approved):

kecnry:
What version/branch of phoebe does this affect (so we can update the merge target to match)?

Raindogjones:
It affects the release branch, looks like it might go all the way back to the 2.2.0 release!

kecnry:
Ok, I retargeted to the bugfix branch.  Can you write a snippet of how you think this should be listed in the changelog for the next bugfix release?  Once 2.4.15 is released, I'll then propagate this through to all other branches.

Raindogjones:
"Fixed population of wavelength array in load function for passbands"?

PR #930, opened 2024-08-27
Impose numpy version cap to 2.0.0.

aprsa:


kecnry:
Let's do this for now - we'll definitely want to track down and fix support for numpy 2.0+ soon, but better to clog the leaks first!  Thanks!

PR #937, opened 2024-09-13
Correction of residuals for interferometry

miroslavbroz:
The reason is that one cannot interpolate in time (many u, v, vis measurements are for the same time).

### DIFF
--- a/phoebe/parameters/parameters.py
+++ b/phoebe/parameters/parameters.py
@@ -3804,7 +3804,7 @@ class ParameterSet(object):
             raise ValueError("model and dataset do not have the same default_unit, cannot interpolate")
 
         # no interpolation for interferometry
-        if qualifier in ['vises', 'clos', 't3s']:
+        if dataset_kind in ['vis', 'clo', 't3']:
             model_interp = model_param.value
             residuals = dataset_param.value - model_param.value
         else:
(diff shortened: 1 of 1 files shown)


phoebe/parameters/parameters.py
```diff
@@ -3803,8 +3803,13 @@ def calculate_residuals(self, model=None, dataset=None, component=None,
         if dataset_param.default_unit != model_param.default_unit:
             raise ValueError("model and dataset do not have the same default_unit, cannot interpolate")
 
-        model_interp = model_param.interp_value(times=times, consider_gaussian_process=consider_gaussian_process)
-        residuals = np.asarray(dataset_param.interp_value(times=times, consider_gaussian_process=consider_gaussian_process) - model_interp)
+        # no interpolation for interferometry
+        if dataset_kind in ['vis', 'clo', 't3']:
+            model_interp = model_param.value
+            residuals = dataset_param.value - model_param.value
```
kecnry, line 3809: this assumes that the model and dataset have the same times.  Since `calculate_residuals` is available to be called directly by the user, we'll probably want to check for that and raise an error if that is not the case.  

For solvers that call this themselves, we probably will want to check in advance as part of `run_checks_solver` to enforce that the input times (https://phoebe-project.org/docs/latest/tutorials/solver_times) will match the dataset times.

miroslavbroz: I do not think this check will ever be needed, because VIS, CLO, T3 datasets are handled in a different way (cf. original_index). One can call calculate_residuals() without problems. The residuals as well as chi^2 is correct (w.r.t. manual computations). No other datasets are affected.

PR #940, opened 2024-09-23
remove legacy phoebe-server

matthiasfabry:
First PR of mine; to remove a (most likely legacy) version of phoebe-server. All functionality should also be present in `client-server/phoebe-server.py`


kecnry (approved):
Thanks!

PR #943, opened 2024-09-26
remove multi support

kecnry:
This PR removes support for multiples from the development branch.  Once this is merged, we'll create a new feature branch that will revert this PR to re-introduce everything there.


aprsa (approved):
As far as I can tell, this is pretty exhaustive. If we run into anything else, we can remove/migrate it there and then.

PR #944, opened 2024-09-26
remove (commented out) vsini/vrot code

kecnry:
Support for vsini / vrot constaints was never finished but was left in the code as comments.  This PR removes that code and can then be moved to a branch in case we want to pick it back up later.

Added a link back here from https://github.com/phoebe-project/phoebe2/discussions/478#discussioncomment-2703734 in case someone wants to pick this up.


aprsa:
Is there a reason why this was commented out in the first place? Glancing at it makes me think it shouldn't take too much effort to get it into a working state? I imagine that misalignment causes problems? Either way, I'm good with having it removed and then re-introduced at a later time if we can't fix it easily.

kecnry:
There's also definition issues that we need to discuss/sort out, especially for vsini.  Agreed though that this is low-hanging fruit for adding back in or even just making as a tutorial for custom constraints.

PR #945, opened 2024-09-27
Correction of constraints

miroslavbroz:
This is formatting of strings in constraints, which is necessary for double precision. Beware that "%.30f" is not enough for tiny numbers! It is safe to use "%.17e" though.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,7 @@ CHANGELOG
 * sigmas_lnf: change handling of noise-nuissance parameter for RVs to no longer depend on the RV amplitude. [#901]
 * Remove duplicated phoebe-server code. [#940]
 * Fix python 3.12+ support by updating invalid escape sequences. [#948]
+* Improved precision in calculation of constraints. [#945]
 
 ### 2.4.14
 
--- a/phoebe/frontend/default_bundles/default_binary.bundle
+++ b/phoebe/frontend/default_bundles/default_binary.bundle
@@ -144,7 +144,7 @@ null
 "kind": "star",
 "context": "component",
 "description": "Critical (maximum) value of the equivalent radius for the given morphology",
-"value": 2.013275176537638,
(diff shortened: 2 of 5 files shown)


kecnry:
Thanks!  

I think we'll also want to rebuild the default binaries (run `frontend/default_bundles/create_default_bundles.py` and then push the changed bundle files) - these are then used by default when users call `phoebe.default_binary()` etc instead of running all the code to generate them each time.

I'm also curious if this adds any extra expense, so maybe we should time the running of constraints before and after this just to make sure the extra cost - if any - is acceptable?

aprsa (approved):

kecnry:
@miroslavbroz - would you be able to do some benchmarking or should I do that?

miroslavbroz:
Could you do it, please? I am not sure what benchmarks do you have in mind. Hmm, I don't expect that changing "%f" -> "%.17e" will affect any benchmarks, because the original code contained even "%.30f".

kecnry:
Benchmarking results:

```
import phoebe
from time import time

b = phoebe.default_binary()

start = time()
for i in range(1000):
    b.set_value(qualifier='requiv', component='primary', value=1.0+i*0.01)
print(time()-start)
```

On this branch (4 different runs): 22.844, 22.588, 22.645, 22.762
Before this branch (4 different runs): 23.341, 22.769, 23.271, 23.061

which to me suggests no noticeable affect (unsurprising, but just wanted to confirm so we don't slip in something that slows phoebe down).

I also rebased and rebuilt the default bundles on this branch so they are tagged correctly (and slight changes to the values - not sure if that's because they were built on a different branch or just floating point precision differences between machines).  If everything still looks good to you, I'll go ahead and approve and merge this in for the 2.4.15 release.

kecnry (approved):
Thanks!  I'll merge once tests are passing and then work on the release.

PR #947, opened 2024-09-30
Dynamics for multis

miroslavbroz:
Re-implementation of Keperian dynamics for N bodies, so that it is compatible with Rebound numerical integrator.


phoebe/dynamics/coord_b2j.py
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env python3
```
aprsa, line 1: Can you please consolidate all source files related to transformations into a single python file, say transformations.py, so that we have everything in one place? Also, could you move executable parts to a test suite so that it's run alongside other tests? Finally, could you write up a high level documentation for the new implementation? It would be good to put our heads together and discuss naming -- for example, 'xyz' vs. 'cartesian' and similar.

PR #948, opened 2024-10-02
Fixes invalid escape sequences

aprsa:
Strings with invalid escape sequences raised a deprecation syntax warning in python 3.11 but a syntax error in 3.12+. All instances (at least the ones that present themselves at import time) have been fixed.


kecnry:
Note that this will be also applied upstream by https://github.com/kecnry/distl/pull/10

PR #959, opened 2024-10-14
phoebe-server updates

kecnry:
This PR updates `phoebe-server` to be functional on modern browser and removes deprecation warnings.  This and https://github.com/phoebe-project/phoebe2-ui/pull/28 will hopefully get the UI in a working state again.

~There still is a remaining issue on main regarding log10 in constraints that seems to be blocking this from working, so will try to fix that before testing this fully.~

Closes https://github.com/phoebe-project/phoebe-development/issues/121


aprsa (approved):
looks good to me!

PR #965, opened 2024-10-18
Fix alignment check

### LINKED ISSUE #773: near-critical system fails Omega_at_vol
To reproduce:

```
import numpy as np
import libphoebe
libphoebe.roche_misaligned_Omega_at_vol(0.011969472802319008, 0.3664, 11.66, 1, np.array([ 7.84353845e-18, -2.39435968e-34,  1.00000000e+00]))
```

setting the spin axis to `[0., 0., 1.]` succeeds and returns `8.93888521685666`.

We can add logic to pass exactly `[1, 0, 0]` instead of computing the spin vector (sounds easy, but that part of the code doesn’t technically have access to the pitch/yaw, so it will take some thinking), but in theory there could also be an intentionally misaligned system with similar inputs so handling these cases in libphoebe might also be beneficial.

horvatm:
I think I corrected the problem connected to issue #773 .

Closes #773

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,8 @@ CHANGELOG
 
 ### 2.4.16
 
+* Fix handling of floating-point precision near the aligned case that used to result in error from libphoebe. [#965]
+
 
 ### 2.4.15
 
--- a/phoebe/lib/bodies.h
+++ b/phoebe/lib/bodies.h
@@ -1,8 +1,8 @@
 #pragma once
 
 /*
-  Examples of implicitely defined closed and connected surfaces in 3D
-  using standard Euclidian coordinate system (x,y,z).
+  Examples of implicitly defined closed and connected surfaces in 3D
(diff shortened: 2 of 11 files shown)


kecnry:
Awesome! 🤞 Looks like this was branched off of `master` but then targets `bugfix-2.4.16` - want me to move the commits and force push to your branch?

horvatm:
Yes, it seems that I did this. I can create another branch from bugfix, if you need this. I just cherry pick the last two commits.

kecnry (approved):
Confirmed that the example from the issue no longer raises an error on my machine and the diff looks reasonable to me.  Thanks! 😄

kecnry:
@horvatm - is "Fix handling of floating-point precision near the aligned case that used to result in error from libphoebe" a good changelog entry for this?  Or do you have suggested language to explain this better?

PR #968, opened 2024-10-23
python 3.13 support

### LINKED ISSUE #967: python 3.13 support
PHOEBE is currently broken under python 3.13, with the first visible error likely being:

```
NameError: Constraint 'logg@primary@star@constraint' raised the following error while attempting to solve for 'logg@primary@star@component'.  Consider flipping the constraint or changing the value of one of ['mass@primary@star@component', 'requiv@primary@star@component', 'logg@primary@star@component'] until the constraint succeeds.  Original error: name 'log10' is not defined
```

We will work towards updating the code to be compatible with python 3.13, but in the meantime suggest downgrading to use python 3.11 or 3.12.

kecnry:
This PR is to test python 3.13 support and eventually make phoebe compatible.

Before merge:
- [x] decide which version of python we want kept in the test matrix (currently only testing 3.13 to see what other errors arise).

Closes #967

### DIFF
--- a/.github/workflows/on_pr.yml
+++ b/.github/workflows/on_pr.yml
@@ -10,8 +10,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.13"]
-        os: [ubuntu-22.04, macos-13, macos-14]  # https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
+        os: [ubuntu-lateset, macos-13, macos-14, macos-latest]  # https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners
         toolchain:
             - {compiler: gcc, version: 13}  # https://github.com/fortran-lang/setup-fortran
 
--- a/.github/workflows/on_pr.yml
+++ b/.github/workflows/on_pr.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.13"]
         os: [ubuntu-22.04, macos-13, macos-14]  # https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners
         toolchain:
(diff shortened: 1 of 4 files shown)


.github/workflows/on_pr.yml
```diff
@@ -10,8 +10,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
-        os: [ubuntu-22.04, macos-13, macos-14]  # https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
+        os: [ubuntu-lateset, macos-13, macos-14, macos-latest]  # https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners
```
kecnry, line 14: @aprsa - does this match what you were suggesting?

aprsa: Yes, but without the typo. ;)

phoebe/parameters/parameters.py
```diff
@@ -12024,7 +12024,8 @@ def _value(var, string_safe_arrays=False, use_distribution=None, needs_builtin=F
                     # to do from builtin import * (and even if I did, python
                     # yells at me for doing that), so instead we'll add them
                     # to the locals dictionary.
-                    locals()[func] = getattr(builtin, func)
+                    # See https://peps.python.org/pep-0667/
+                    sys._getframe().f_locals[func] = getattr(builtin, func)
```
aprsa, line 12028: do you know perhaps which is the earliest version of python for which this will work?

README.md
```diff
@@ -88,6 +88,7 @@ CHANGELOG
 ### 2.4.16
 
 * Fix handling of floating-point precision near the aligned case that used to result in error from libphoebe. [#965]
+* Fix support for python 3.13, remove official support for python 3.8. [#968]
```
aprsa, line 91: you're saying remove 3.8 here, but in the toml file we have >=3.8; one of them needs to be updated.

kecnry: ah, good catch, I meant removing support for 3.7.  I'll update that here.

phoebe/parameters/parameters.py
kecnry: no, but it passes on the full CI matrix (3.8+).

aprsa (approved):

PR #974, opened 2024-10-28
Fixes list_passband_online_history()

aprsa:
The call to this function expected to receive a dict from the server but it instead received a list. The list is now converted to a dict. It does so by parsing a fixed number of characters to determine a date string, which might be problematic for non-EN locales but I can't think of any robust way to make this work automatically.


kecnry:
Where did this break?  Does it work in the released version?

aprsa:
This issue arose as a result of the underlying change to how history keywords in fits files are parsed. Thus, it is better that we fix it there instead of creating this workaround.

PR #979, opened 2024-11-07
Fixes passband history adding/parsing

aprsa:
In previous versions, history was overloading the HISTORY keyword in fits files and implementing a custom -END- separator, thus defeating the purpose of the fits standard. This has now been replaced with multiple HISTORY entries in the fits files.

Also, the timestamps thus far were in the ctime() format, which is locale-dependent. The new format follows the ISO standard and is locale-independent.

All parsers are implemented to use either format.

This version is currently serving passbands on staging.phoebe-project.org and everything is working as expected. Thus, I propose to pull this into the blending feature branch and use it to serve the passbands from the official tables.phoebe-project.org API.

### DIFF
--- a/phoebe/atmospheres/passbands.py
+++ b/phoebe/atmospheres/passbands.py
@@ -353,7 +353,7 @@ class Passband:
         if len(comment) > max_length:
             raise ValueError(f'comment length should not exceed {max_length} characters.')
 
-        self.comments.append(comment)                
+        self.comments.append(comment)
 
     def on_updated_ptf(self, ptf, wlunits=u.AA, oversampling=1, ptf_order=3):
         """
--- a/phoebe/atmospheres/passbands.py
+++ b/phoebe/atmospheres/passbands.py
@@ -2,7 +2,6 @@ from phoebe import __version__ as phoebe_version
 from phoebe import conf, mpi
 from phoebe.utils import _bytes
 from tqdm import tqdm
-from itertools import product
 
 import ndpolator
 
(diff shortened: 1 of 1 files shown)


phoebe/atmospheres/passbands.py
```diff
@@ -38,7 +36,7 @@
 
 # define the URL to query for online passbands.  See tables.phoebe-project.org
 # repo for the source-code of the server
-_url_tables_server = 'http://tables.phoebe-project.org'
+_url_tables_server = 'https://staging.phoebe-project.org'
```
kecnry, line 39: this definitely shouldn't be merged, right?

phoebe/atmospheres/passbands.py
```diff
@@ -305,33 +303,57 @@ def log(self):
...
+        >>> pb.add_comment('this is a comment.')
         """
 
         if not isinstance(comment, str):
             raise ValueError('passband header comments must be strings.')
+        if len(comment) > max_length:
+            raise ValueError(f'comment length should not exceed {max_length} characters.')
 
-        self.comments.append(comment)
+        self.comments.append(comment)
```
kecnry, line 356:
```suggestion
        self.comments.append(comment)
```

phoebe/atmospheres/passbands.py
aprsa: It should be merged into feature-blending. Staging is running the blending version of phoebe within the server.

PR #952, opened 2024-10-03
modified publish workflow

kecnry:
This replaces #949, and now adapts the matrix wheel generation used by https://github.com/aprsa/ndpolator/blob/main/.github/workflows/on_release.yaml

PR #982, opened 2024-11-13
Numpy2

mwrona77:
This is my first pull request on GitHub, so if I did something wrong, consider it a rookie mistake. 

Since part of my issue involved changing `np.product` to `np.prod`, I made that adjustment separately from @giammarc.

Although this change was enough for all the tests to pass, there were a lot of warnings, mostly about the deprecation of certain functions. All other code changes I made are specifically to eliminate these warnings.

I understand that the next step is CI, but I assume that to run the tests, I need to make this pull request. Or should I have done this earlier?

Closes https://github.com/phoebe-project/phoebe-development/issues/110


phoebe/dependencies/ligeor/models/polyfit.py
```diff
@@ -1,7 +1,7 @@
 import numpy as np
 from operator import itemgetter
 from itertools import groupby
-from numpy.core.fromnumeric import mean
+from numpy import mean
```
kecnry, line 4: would be great to also make this suggestion upstream to https://github.com/gecheline/ligeor

phoebe/frontend/io.py
```diff
@@ -465,8 +465,16 @@ def load_legacy(filename, add_compute_legacy=True, add_compute_phoebe=True,
...
-                        converters={0: lambda s: s.strip(), 1: lambda s: s.strip()})
+    params = np.genfromtxt(
+        filename,
+        dtype='str',
+        delimiter='=',
+        converters={0: lambda s: s.strip().decode() if isinstance(s, bytes) else s.strip(),
+                    1: lambda s: s.strip().decode() if isinstance(s, bytes) else s.strip()},
+        comments='#',       
+        filling_values='',   
+        autostrip=True
```
kecnry, line 476: does `autostrip` remove the need for `.strip` in the converter?

kecnry (commented):
The diff looks reasonable to me - thanks!  For some reason there is one build that is still failing - any idea why that is?

phoebe/dependencies/ligeor/models/polyfit.py
mwrona77: Sure, but in what form? A pull request, or just a mention somewhere?

It could also be changed to `from numpy._core.fromnumeric import mean` (see underline before `core`),  but that won't work with NumPy < 2.0.

`from numpy import mean` essentially does the same thing and works in both cases.

kecnry: Yeah, just a quick PR to that repo so the two are in sync.  Eventually we'll probably move away from packaging these within phoebe and instead have pip pull them in, so we don't want changes in phoebe that don't exist in the actual packages themselves.

phoebe/frontend/io.py
mwrona77: Yes, you're right. I just tested three combinations:
- without `autostrip = True`
- without `strip()` in converters
- without converters

All of them passed the tests for Python 3.9 and 3.10. I'll leave it up to you to decide which should I keep. The most convenient is probably the one without `converters` and with `autostrip=True`, but the one closest to the original is without `autostrip = True`.

mwrona77:
> For some reason there is one build that is still failing - any idea why that is?

I'm not entirely sure, but there's a high chance that NumPy no longer supports building i686 wheels. See, for example: https://github.com/duckdb/duckdb/pull/9101

This would make sense, as the CI fails specifically at 
`Building pp310-manylinux_i686 wheel`
`PyPy 3.10 manylinux i686`.

Honestly, I have no idea how to resolve this. One option is to enforce using NumPy < 2.0 when building wheels for i686, or to drop support for i686 entirely.

Here's what I found in NumPy's `pyproject.toml` on GitHub (https://github.com/numpy/numpy/blob/060c28aa1920f0699ef7c3725e59dcd8e99723ca/pyproject.toml#L143):
```
[tool.cibuildwheel]
# Note: the below skip command doesn't do much currently, the platforms to
# build wheels for in CI are controlled in `.github/workflows/wheels.yml` and
# `tools/ci/cirrus_wheels.yml`.
build-frontend = "build"
skip = "cp36-* cp37-* cp-38* pp37-* *-manylinux_i686 *_ppc64le *_s390x *_universal2"
before-build = "bash {project}/tools/wheels/cibw_before_build.sh {project}"
# The build will use openblas64 everywhere, except on arm64 macOS >=14.0 (uses Accelerate)
config-settings = "setup-args=-Duse-ilp64=true setup-args=-Dallow-noblas=false build-dir=build"
before-test = "pip install -r {project}/requirements/test_requirements.txt"
test-command = "bash {project}/tools/wheels/cibw_test_command.sh {project}"
```
`*-manylinux_i686 *` is in `skip`

phoebe/dependencies/ligeor/models/polyfit.py
mwrona77: Done - [https://github.com/gecheline/ligeor/pull/1](url)

mwrona77:
> ok, then I'm ok with just adding to skip as well in this case. Anyone with that scenario can try to build from source as we used to do before #952

Done! This time it passes all the tests. I added `*-manylinux_i686` in the `publish.yml` to the `CIBW_SKIP` line.

kecnry (approved):
LGTM - thanks for the fix!

PR #983, opened 2024-11-14
Adds Tremblay atmospheres

Raindogjones:


aprsa (approved):
This looks good to me. I want to relocate magic strings to the top of the module (so that we never actually refer to individual atmospheres anywhere in the code), but this needs to be done systematically and is outside of the scope for this PR.

aprsa:
Are Tremblay tables also on Terra?

Raindogjones:
Yep, the 4 TMAP grids and Tremblay are all tarballed in my home directory.

phoebe/frontend/default_bundles/default_binary.bundle
```diff
@@ -426,9 +426,13 @@ null
...
 "choices": [
 "auto",
+"tmap_DA",
 "phoenix",
 "tmap_sdO",
-"ck2004"
+"tmap_DO",
+"ck2004",
+"tremblay",
+"tmap_DAO"
```
kecnry, line 435: were the tmap ones just overlooked in the past?

phoebe/frontend/default_bundles/default_contact_binary.bundle
```diff
@@ -144,7 +144,7 @@ null
 "kind": "star",
 "context": "component",
 "description": "Critical (maximum) value of the equivalent radius for the given morphology",
-"value": 1.6724563972838287,
+"value": 1.6724563972838289,
```
kecnry, line 147: I vaguely remember seeing updates here recently... do we understand why this is changing?

phoebe/frontend/default_bundles/default_binary.bundle
Raindogjones: I think we just never rebuilt the default bundles in this particular branch.

PR #1017, opened 2025-03-17
singlestar/misaligned spots fixes

### LINKED ISSUE #1008: Single rotating star is not rotating
Hello,

I'm trying to use Phoebe 2.4.17 to model the light curve of a single rotating star with a spot. Unfortunately, I found out that the output light curve is completely flat because Phoebe does not make the star rotate, even though I properly set the rotation period. I was able to reproduce the same error using this tutorial: https://phoebe-project.org/docs/2.4/examples/single_spots. Would you be so kind as to help me solve this issue?

Thanks in advance!

MichaelAbdul-Masih:
This pull request includes fixes for two different bugs related to spots.  In particular for the misaligned spots fix, I would appreciate it if someone could double check that the spot placement is working as expected. 

For a bit of context, the `pointing_vector` function within the `Spot` feature calculates the location of the spot on the mesh which requires being in the star's local frame.  The issue appears to be in how this coordinate transform was calculated and the fact that the pointing vector was not renormalized. I've corrected this by instead using cross products to determine the coordinate system of the star and then divided the final pointing vector by its length.

I would appreaciate a second set of eyes (and brain) to make sure this is being calculated correctly.

Fixes #1008


phoebe/backend/universe.py
```diff
@@ -2432,7 +2435,7 @@ def is_convex(self):
     def needs_recompute_instantaneous(self):
         # recompute instantaneous for asynchronous spots, even if meshing
         # doesn't need to be recomputed
-        return self.needs_remesh or (not self.is_single and len(self.features) and self.F != 1)
+        return self.needs_remesh or (len(self.features) and self.F != 1.0) or self.is_single
```
aprsa, line 2438: won't this force recomputing whenever self.is_single is True? Why is this necessary?

phoebe/backend/universe.py
```diff
@@ -3167,19 +3170,36 @@ def pointing_vector(self, s, time):
...
-        # define the basis vectors in the spin (primed) coordinates in terms of
-        # the Roche coordinates.
-        # ez' = s
-        # ex' =  (ex - s(s.ex)) /|i - s(s.ex)|
-        # ey' = s x ex'
+        '''
+        define the basis vectors in the spin (primed) coordinates in terms of
+        the Roche coordinates.
+        - the z' direction is explicitly given by the spin vector
+        ez' = s
```
aprsa, line 3177: is it safe to assume that `s` is a unit vector?

phoebe/backend/universe.py
```diff
@@ -3167,19 +3170,36 @@ def pointing_vector(self, s, time):
...
+        that the y' direction should always be orthogonal to x. Thus, to get this,
+        we'll first calculate the y' direction by taking the cross product of the 
+        z' and x directions 
+        ey' = ez' x ex
+        if the pole aligns with the x-axis, then the y' direction is the z direction
+        - the x' direction is then defined by the cross product of the other two
+        ex' = ey' x ez'
+        '''
         ex = np.array([1., 0., 0.])
         ezp = s
```
aprsa, line 3188: should we divide by |s| for the peace of mind?

phoebe/backend/universe.py
```diff
@@ -3167,19 +3170,36 @@ def pointing_vector(self, s, time):
...
-        # the Roche coordinates.
-        # ez' = s
-        # ex' =  (ex - s(s.ex)) /|i - s(s.ex)|
-        # ey' = s x ex'
+        '''
+        define the basis vectors in the spin (primed) coordinates in terms of
+        the Roche coordinates.
+        - the z' direction is explicitly given by the spin vector
+        ez' = s
+        - the x' direction is defined as being towards the companion, which means
```
aprsa, line 3178: this should be x, not x', right?

phoebe/backend/universe.py
```diff
@@ -3167,19 +3170,36 @@ def pointing_vector(self, s, time):
...
+        exp = np.cross(eyp, ezp)
 
-        return np.sin(self._colat)*np.cos(longitude)*exp +\
+        # now we can express the pointing vector in terms of the primed basis
+        pv = np.sin(self._colat)*np.cos(longitude)*exp +\
                   np.sin(self._colat)*np.sin(longitude)*eyp +\
                   np.cos(self._colat)*ezp
+        
+        # renormalize and return pointing vector
+        return pv / np.sqrt(np.sum(pv**2))
```
aprsa, line 3201: there's np.linalg.norm() for that. But also, if `s` is normalized, `pv` should also always be normalized.

aprsa (commented):
Looks good to me! I left a few comments that I'd love to get your input on, but otherwise I'd say this is good to go!

phoebe/backend/universe.py
kecnry: we definitely want to only do this when absolutely necessary

phoebe/backend/universe.py
MichaelAbdul-Masih: No, here I do mean x', but perhaps I should rephrase this a bit.  What I mean is that the x' direction should point in the longitudinal direction that the companion is in the rotating frame.

phoebe/backend/universe.py
MichaelAbdul-Masih: I ran some tests and it appears to be, but as you mentioned in the comment below, I'll renormalize it explicitly just in case

phoebe/backend/universe.py
MichaelAbdul-Masih: yep good idea

phoebe/backend/universe.py
MichaelAbdul-Masih: I'll switch to np.linalg.norm, but I'll keep the renormalization just in case.

phoebe/backend/universe.py
```diff
@@ -2432,7 +2435,7 @@ def is_convex(self):
     def needs_recompute_instantaneous(self):
         # recompute instantaneous for asynchronous spots, even if meshing
         # doesn't need to be recomputed
-        return self.needs_remesh or (not self.is_single and len(self.features) and self.F != 1)
+        return self.needs_remesh or (len(self.features) and self.F != 1.0) or (len(self.features) and self.is_single)
```
kecnry, line 2438: This just reads a little more intuitively to me 🤷‍♂️   The comment above probably should be updated as well to explain why the single case needs recomputing.
```suggestion
        return self.needs_remesh or (len(self.features) and (self.is_single or self.F != 1.0))
```

kecnry:
Changes look reasonable to me as long as tests pass 🤞  Can you add a changelog entry (or two if that's easier) please?

README.md
```diff
@@ -80,6 +80,8 @@ CHANGELOG
 
 ### 2.4.18
 
+* Fix handling of spots in single star rotstar case where spots were not co-rotating properly [#1008]
+* Fix misalined spots bug that caused size of spot to change across the rotation period
```
kecnry, line 84:
```suggestion
* Fix handling of spots in single star rotstar case where spots were not co-rotating properly [#1017]
* Fix misalined spots bug that caused size of spot to change across the rotation period [#1017]
```

README.md
```diff
@@ -80,6 +80,8 @@ CHANGELOG
 
 ### 2.4.18
 
+* Fix handling of spots in single star rotstar case where spots were not co-rotating properly [#1017]
+* Fix misalined spots bug that caused size of spot to change across the rotation period [#1017]
```
kecnry, line 84: Sorry missed this one earlier!

```suggestion
* Fix misaligned spots bug that caused size of spot to change across the rotation period [#1017]
```

kecnry (approved):

Issue #754, opened 2023-06-21
passing nparray to animate times fails


kecnry:
but casting to list/numpy array is successful

PR #1018, opened 2025-03-17
Animation times bugfix

MichaelAbdul-Masih:
fixes issue #754


kecnry:
so was the problem ultimately in comparing a single element in `kwargs.values()` to the nparray object because it was checking with `==` or `is` on the element?  I wonder if we could/should update nparray to handle the comparison correctly instead (or in addition)?

PR #1020, opened 2025-03-26
Contact envelope reflection optimization

matthiasfabry:
Currently, `libphoebe.mesh_radiosity_problem_nbody_convex()` was called each iteration, leading to un-needed slowdown.  
I've overridden the appropriate properties of `Envelope` so this does not happen. In the current implementation of contact envelopes, I do not think we need to support any time/phase dependent quantities.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,10 @@ To understand how to use PHOEBE, please consult the [tutorials, scripts and manu
 CHANGELOG
 ----------
 
+### 2.4.18
+
+
+
 ### 2.4.17
 
 * Fix support for numpy 2.0. [#982]
--- a/phoebe/__init__.py
+++ b/phoebe/__init__.py
@@ -17,7 +17,7 @@ Available environment variables:
 
 """
 
-__version__ = '2.4.17'
+__version__ = '2.4.18'
 
 import os as _os
 import sys as _sys
--- a/phoebe/backend/universe.py
+++ b/phoebe/backend/universe.py
@@ -2888,6 +2888,21 @@ class Envelope(Body):
     def mesh(self):
         return self.meshes
 
+    @property
+    def needs_recompute_instantaneous(self):
+        """
+        contact envelopes never need quantities recomputed (we assume steady state envelopes). Only line of sight to
+        the binary matters
+        """
+        return False
+
+    @property
+    def needs_remesh(self):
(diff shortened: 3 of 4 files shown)


matthiasfabry:
Ok, given the errors in 'contact' and 'reflection', I don't seem to understand how reflection actually is computed. reverting to draft

matthiasfabry:
I've been looking at this in a bit more detail and I do think I am correct in that reflection should only be computed once for (circular, steady state) contact binaries.

The test case fails because reflection is not applied on any point after the first (it's subtle to see on the figure but it is there):
 Screenshot 2025-04-01 at 4 13 20 PM

I believe this has to do with copying over the post-reflected teffs to the appropriate meshes.
`universe.py:420` has the following section after computing the radiosity problem:
```
        if not self.needs_recompute_instantaneous:
            logger.debug("reflection: copying updated teffs to standard mesh")
            theta = 0.0
            standard_meshes = mesh.Meshes({body.component: body._standard_meshes[theta] for body in self.mesh_bodies})
            standard_meshes.set_column_flat('teffs', teffs_intrins_and_refl_flat)

            self.is_first_refl_iteration = False
```
I'm not yet familiar enough with how the contact meshes/halves are handled internally, but could it be that an exception needs to be written somewhere in here?
Another possibility is that the contact envelope mesh is not built from the appropriate meshes when using saved standard meshes (in steps after the first).

Any ideas?

kecnry:
Its definitely possible that that's the case and the tests are incorrect.  Can you maybe expose the mesh at two times (perhaps even the same phase but one cycle apart) with and without irradiation enabled to show if this is indeed what is happening?

matthiasfabry:
At times 0 and 1.5 (both phase 0).
With (wilson) irradiation:
Screenshot 2025-04-02 at 1 19 13 PM
Without irradiation:
Screenshot 2025-04-02 at 1 22 16 PM

Seems pretty clear to me that the initial mesh is "forgotten."

kecnry:
rebasing on top of `bugfix-2.4.22` should fix the jktebop test failures and let the rest of the tests run

kecnry:
is this intended to be a bugfix or no?  I'll need to update the feature branch to get the jktebop fix yet if you want it rebased on that

PR #1021, opened 2025-03-27
Feature mf marching

MichaelAbdul-Masih:
multi-frontal marching


phoebe/distortions/rotstar.py
```diff
@@ -24,12 +24,15 @@ def rotfreq_to_omega(rotfreq, M_star, scale=c.R_sun.si.value, solar_units=False)
...
     """
+
     if solar_units:
         omega = rotfreq / np.sqrt(M_star * c.G.to(u.solRad**3/(u.solMass*u.d**2)).value/scale**3)
     else:
         # then SI units
         omega = rotfreq / np.sqrt(M_star * c.G.value/scale**3)
 
+    # ~ print("*** rotstar.rotfreq_to_omega", rotfreq, scale, solar_units, omega)
+
```
kecnry, line 35:
```suggestion
```

phoebe/lib/bodies.h
```diff
@@ -96,7 +96,7 @@ struct Ttorus {
 /* ===================================================================
   Sphere
 
-  Defined of implicitly by a constrain
+  Defined of implicitly by a constraintt
```
kecnry, line 99:
```suggestion
  Defined of implicitly by a constraint
```

kecnry (commented):
Thanks for rebasing this - this really will be great to get in!  Once segfaults are fixed, we might want to run some manual testing if we want to get this as a bugfix (which would be nice) in case something isn't caught by existing test coverage.  Alternatively, we probably could merge quicker into `release-2.5`... but that also wouldn't get released as soon.

phoebe/lib/bodies.h
aprsa: Can anyone reproduce the segfault? It looks to be isolated to macs. I'd be happy to look at the trace if anyone can reproduce.

aprsa: ... also, @horvatm, do you perhaps have any ideas on what might be causing the segfault?

horvatm: I very confused by this version. I have surely introduced several helper functions to make finding correct index on the front_poligon robust already last year, for the workshop.

kecnry: @horvatm - do you know where that code lives?

horvatm: I think these bugs connected to mac and marching have been corrected in branch fix_alignment_check (https://github.com/horvatm/phoebe2-fork)  about 10 month ago, or at least they seem to be very similar.

Raindogjones: I can reproduce. Here is the output from the logger at level='DEBUG' as well as the system report returned from the segfault.
[log.log](https://github.com/user-attachments/files/19582637/log.log)
[report.txt](https://github.com/user-attachments/files/19582638/report.txt)

MichaelAbdul-Masih: @horvatm , indeed, this is from the old branch from 2021 just rebased up to 2.4.17 so its possible there have been changes since then in other branches that take care of this issue.  I'll look into the fix_alignment_check branch and see if merging in those changes fixes these bugs.  Thanks for the suggestion!

@Raindogjones thanks for reproducing!

PR #1037, opened 2025-05-08
Adds dockerizing capability.

aprsa:
This Dockerfile builds a base (parent) image that can be used for streamlined inheritance for any app that uses phoebe under the hood. It supports deploying containers for different branches. To build an image, do:

```bash
docker build --build-arg PHOEBE_BRANCH=master -f Dockerfile -t phoebe2:main .
```

Replace the branch name and the tag appropriately for other branches. To test an image, do:

```bash
docker run --rm phoebe2:main
```

This will print the python version and the phoebe version and exit.

To reiterate, this docker image is not meant to be used as a standalone service but as a base (parent) image for other containers, for example the tables server or the UI or anything else that requires phoebe.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,22 @@
+FROM python:3-alpine
+
+ARG PHOEBE_BRANCH=master
+
+RUN apk add --no-cache \
+      git \
+      build-base \
+      musl-dev
+
+RUN git clone --branch ${PHOEBE_BRANCH} --single-branch --depth 1 \
+      https://github.com/phoebe-project/phoebe2.git /phoebe
+
+WORKDIR /phoebe
+
+RUN pip install --upgrade pip
+
+RUN pip install --no-cache-dir .
+
+RUN mkdir -p /workspace
+WORKDIR /workspace
+
+ENTRYPOINT ["python", "-c", "import sys, phoebe; print(f'Python {sys.version}\\nPHOEBE {phoebe.__version__}')"]


kecnry:
Is this needed in the bugfix or can it wait until 2.5?

aprsa:
I will use it rightaway for deployment on terra, which means that I'll need to keep the files backported from 2.5 if we defer. Given it's disjunct from everything else, I figured it wouldn't hurt to sneak it into the bugfix.

That said, I'm fine either way, given that 2.5 now really shouldn't be too far away.

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+ARG PHOEBE_BRANCH=master
+
+RUN apk add --no-cache \
+      git \
+      build-base \
+      musl-dev
+
+RUN git clone --branch ${PHOEBE_BRANCH} --single-branch --depth 1 \
+      https://github.com/phoebe-project/phoebe2.git /phoebe
```
kecnry, line 11: doesn't running this file require to already have cloned the repo though?  Would it make more sense to either keep this external or ship it with phoebe but have the branch handling done via git outside of this?

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+WORKDIR /phoebe
+
+RUN pip install --upgrade pip
+
+RUN pip install --no-cache-dir .
+
+RUN mkdir -p /workspace
+WORKDIR /workspace
```
kecnry, line 20: is this useful (or potentially harmful/unexpected) to anyone else or just the specific setup for the web servers?  If just us, maybe it should live in one of the server repos instead?

Dockerfile
aprsa: This clones the latest version of the repo branch *inside* the docker image. We could copy the local repo but then you would need to switch branches manually and any local changes might trickle in. I think this is cleaner.

Dockerfile
aprsa: this is just to avoid phoebe being imported from the source tree for reporting python/phoebe versions.

aprsa: The main reason for the base image is inheritance -- any image that requires phoebe will build from this image instead of having to create a fresh image, install all dependencies and build phoebe over and over again. This way people would simply do:
```
FROM phoebe2:blending
. . .
```
and be done. Eventually we may want to host this image on either dockerhub or github container hub.

PR #1033, opened 2025-05-05
Simple boosting test

Raindogjones:
Simple test that checks boosting can be set for one component and that it actually boosts the light curve.

PR #1043, opened 2025-05-29
fix support for astropy 7 unit changes

kecnry:
This PR fixes the astropy unit issues preventing creating a default binary with astropy 7.x seen in https://github.com/phoebe-project/phoebe2/pull/1033 and reported in https://github.com/phoebe-project/phoebe2/discussions/1042

These changes should then also be pushed back to the `distl` and `unitsiau2015` repos.


Raindogjones:
Does this bump the astropy version requirement too or is it backwards compatible?

kecnry:
this should be fully backwards compatible, just _adding_ support for astropy 7 on top of the previous support.

PR #1060, opened 2025-08-03
Environment variable for PHOEBE_TABLES_SERVER

kecnry:
this implements a `PHOEBE_TABLES_SERVER` environment variable to allow overriding the default URL to access the tables server.

### DIFF
--- a/phoebe/atmospheres/passbands.py
+++ b/phoebe/atmospheres/passbands.py
@@ -39,7 +39,7 @@ logger.addHandler(logging.NullHandler())
 
 # define the URL to query for online passbands.  See tables.phoebe-project.org
 # repo for the source-code of the server
-_url_tables_server = 'http://tables.phoebe-project.org'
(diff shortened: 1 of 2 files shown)


aprsa (approved):

phoebe/atmospheres/passbands.py
```diff
@@ -39,7 +39,7 @@
 
 # define the URL to query for online passbands.  See tables.phoebe-project.org
 # repo for the source-code of the server
-_url_tables_server = 'http://tables.phoebe-project.org'
+_url_tables_server = os.getenv('PHOEBE_TABLES_SERVER', 'http://tables.phoebe-project.org')
```
aprsa, line 42: let's replace http with https while we're doing this

aprsa:
I tested and it works as expected.

phoebe/atmospheres/passbands.py
kecnry: there are a lot of other instances, I'll do that across the board in a follow-up PR

PR #1062, opened 2025-08-05
replace http with https where possible

kecnry:


phoebe/atmospheres/passbands.py
```diff
@@ -2816,7 +2816,7 @@ def update_passband(passband, local=True, content=None, gzipped=None):
     <phoebe.atmospheres.passbands.update_passband>.
 
     Download and install updates for a single passband from
-    http://tables.phoebe-project.org, retrieving
+    https://tables.phoebe-project.org, retrieving
```
aprsa, line 2819: should we replace this with "passband server" or something like that because the url (while the default) can be configured?

phoebe/atmospheres/passbands.py
```diff
@@ -2884,7 +2884,7 @@ def update_all_passbands(local=True, content=None):
     <phoebe.atmospheres.passbands.update_all_passbands>.
 
     Download and install updates for all passbands from
-    http://tables.phoebe-project.org, retrieving
+    https://tables.phoebe-project.org, retrieving
```
aprsa, line 2887: same as above

phoebe/atmospheres/passbands.py
```diff
@@ -3032,7 +3032,7 @@ def list_online_passbands(refresh=False, full_dict=False, skip_keys=[], repeat_e
     <phoebe.atmospheres.passbands.list_online_passbands>.
 
     List all passbands available for download from
-    http://tables.phoebe-project.org.
+    https://tables.phoebe-project.org.
```
aprsa, line 3035: same as above

phoebe/atmospheres/passbands.py
```diff
@@ -3127,7 +3127,7 @@ def get_passband(passband, content=None, reload=False, update_if_necessary=False
     will be downloaded and installed locally.  If the installed passband does
     not have the necessary tables to match `content` then an attempt will be
     made to download the necessary additional tables from
-    http://tables.phoebe-project.org
+    https://tables.phoebe-project.org
```
aprsa, line 3130: same as above

phoebe/atmospheres/passbands.py
```diff
@@ -2575,7 +2575,7 @@ def download_passband(passband, content=None, local=True, gzipped=None):
     <phoebe.atmospheres.passbands.download_passband>.
 
     Download and install a given passband from
-    http://tables.phoebe-project.org.
+    https://tables.phoebe-project.org.
```
aprsa, line 2578: see below for the comment, I inadvertently missed this one on first pass

PR #1063, opened 2025-08-05
Catch ellc OSError

kecnry:
Catch any `OSError` from `ellc` and treat as an `ImportError` to address systems where users have installed a broken version of ellc, see https://github.com/pmaxted/ellc/issues/21

### DIFF
--- a/phoebe/backend/backends.py
+++ b/phoebe/backend/backends.py
@@ -45,7 +45,7 @@ else:
 
 try:
     import ellc
-except (ImportError, OSError):
+except Exception:
     _use_ellc = False
 else:
     _use_ellc = True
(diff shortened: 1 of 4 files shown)


phoebe/backend/backends.py
```diff
@@ -45,7 +45,7 @@
 
 try:
     import ellc
-except ImportError:
+except (ImportError, OSError):
```
aprsa, line 48: is it really important to catch particular exceptions? If it fails to import, we don't really care _why_ it failed, so I'd suggest triggering on _all_ exceptions here (and elsewhere in the code).

kecnry: ok, switched to `except Exception` to catch anything

PR #1077, opened 2025-09-05
Energy transfer for contact binaries

matthiasfabry:
Implements parametric contact-binary energy transfer.
4 schemes are currently supported:

1. Simple smoothing of the neck temperatures to avoid discontinuities (implemented by @aprsa, @gecheline)
2. Isotropic mixing where the temperatures depend on the distance to the neck (also implemented by @aprsa, @gecheline)
3. lateral mixing on a horizontal band (designed by @aprsa and @gecheline, improved by me)
4. Perfect mixing, setting temperatures to that of the primary


matthiasfabry:
Clearly I did the rebase incorrectly to fix the conflicts. Would a merge commit been better here? Or did I rebase incorrectly?

PR #1087, opened 2025-10-03
Dependabot

aprsa:
Introduces dependabot script to keep the dependencies (github scripts and packages) current.


aprsa:
> Any way to know if this is configured correctly without merging?

I don't think so -- dependabot only runs on main.

PR #1100, opened 2026-02-02
TMAP/loghefrac support

Raindogjones:
Includes:
#1083 
#1097 
#1027 
#1093


phoebe/atmospheres/passbands.py
```diff
@@ -716,7 +721,7 @@ def load(cls, archive, load_content=True):
                 stored_atms = set([content.split(':')[0] for content in self.content])
 
                 # TODO: replace with < parse('2.5.0') when 2.5.0 is released
-                if parse(self.phoebe_version) != parse('2.4.17.dev+feature-blending'):
+                if parse(self.phoebe_version) != parse('2.4.21.dev+feature-blending'):
```
aprsa: this should be 2.4.22.dev+feature-blending now that 2.4.22 is out

phoebe/atmospheres/models.py
```diff
@@ -74,6 +74,8 @@ class provides a common interface for handling different model
 
     # default axes:
     basic_axis_names = ['teffs', 'loggs', 'abuns']
+    # assumed/fixed axes and their values:
+    fixed_axis_values = {}
```
aprsa: so these are single value axes? Can we call them that, then? single_value_axes?

phoebe/atmospheres/models.py
```diff
@@ -225,6 +243,7 @@ class WDBlackbodyModelAtmosphere(ModelAtmosphere):
 
     name = 'extern_planckint'
     basic_axis_names = ['teffs']
+    fixed_axis_values = {'abuns': 0.0}
```
aprsa: technically, blackbody doesn't have any abuns; do we need this? Do we need loggs as well then?

phoebe/atmospheres/models.py
```diff
@@ -250,6 +269,7 @@ class BlackbodyModelAtmosphere(ModelAtmosphere):
     prefix = 'bb'
 
     basic_axis_names = ['teffs']
+    fixed_axis_values = {'abuns': 0.0}
```
aprsa: same here -- do we need abuns and, if so, do we need loggs?

phoebe/atmospheres/passbands.py
```diff
@@ -883,7 +888,58 @@ def _planck_spi(self, lam, Teff):
         expterm = np.exp(hclkt)
         return hclkt * expterm/(expterm-1)
 
+    def get_axis_limits(self, atm, axis):
```
aprsa: what is this used for? Either way, I'd say this belongs in models.py, not here.

phoebe/backend/mesh.py
```diff
@@ -1069,6 +1070,17 @@ def abuns(self):
         """
         return self._abuns
 
+    @property
+    def loghefracs(self):
+        """
+        Return the array of loghefracs, where each item is a scalar/float.
+
+        TODO: UNIT?
```
aprsa, line 1078: should be unitless, no?

phoebe/backend/universe.py
```diff
@@ -1643,6 +1655,24 @@ def _fill_abuns(self, mesh=None, abun=0.0):
             theta = 0.0
             self._standard_meshes[theta].update_columns(abuns=abun)
 
+    def _fill_loghefracs(self, mesh=None, loghefrac=0.0):
+        """
+        TODO: add documentation
```
aprsa, line 1662: finish TODOs before the merge?

phoebe/frontend/bundle.py
```diff
@@ -4024,6 +4035,45 @@ def ld_coeffs_len(ld_func, ld_coeffs):
                                     self.filter(qualifier='pblum_method', compute=compute, context='compute', **_skip_filter_checks),
                                     True, 'run_compute')
 
+                # check abun and loghefrac values are consistent with the atmosphere
```
aprsa, line 4038: do we have a unit test for this? To make sure that column names are parsed correctly and consistently?

phoebe/frontend/bundle.py
```diff
@@ -10276,16 +10373,25 @@ def compute_ld_coeffs(self, compute=None, set_value=False, **kwargs):
...
 
-                # TODO: generalize this.
-                query_cols = ('teffs', 'loggs', 'abuns')
-                query_pts = np.array(((teff, logg, abun),))
-                query = passbands.InterpQuery(cols=query_cols, pts=query_pts)
+                # TODO: generalize this further (some have neither and so should be able to handle not passing either)
+                ldatm = models._atmtable[ldcs]
+                if ldatm.has_axis('loghefracs'):
+                    loghefrac = self.get_value(qualifier='loghefrac', component=ldcs_param.component, context='component', **_skip_filter_checks)
+                    # TODO: loghefrac should be pluralized!
```
aprsa, line 10383: done, no?

aprsa (commented):
it looks like the biggest change (in terms of models/passbands) is the addition of fixed_value_axis. I wonder whether this is even necessary -- when I first implemented this, I added a `meta` argument to InterpQuery that was meant to do precisely this: pass "virtual" parameters to ndp and ndp-related code.

So the rationale, the way I understand it, is that we pass either abun or loghefrac, depending on the model atm, and if it's a single supported value, the user is still supposed to provide it and phoebe merely checks against fixed_value_axis under the hood?

phoebe/backend/mesh.py
Raindogjones: Yeah, unitless.

phoebe/atmospheres/models.py
kecnry: they hold the values, so I think values should be last, but I'm find with single over fixed

phoebe/atmospheres/models.py
kecnry: we need to know the assumed value for frontend checks _or_ we need to write an exception for blackbody

phoebe/atmospheres/passbands.py
kecnry: yep, this likely wouldn't be caught by rebase because its a new addition in this branch

phoebe/atmospheres/models.py
aprsa: there are no assumed values for loggs/abuns -- blackbody atms depend only on teff.

phoebe/atmospheres/passbands.py
kecnry: this is used in compute-checks to make sure abun/loghefrac are within bounds for the given atmosphere (and could be extended in the future to other values, although we don't have access to per-element values like local temperatures at this point in the code, so it wouldn't completely prevent out-of-bounds issues)

phoebe/atmospheres/passbands.py
Raindogjones: Done

phoebe/atmospheres/models.py
kecnry: (as discussed offline) - then we may want to either store this as `None` or just check if its in basic_axis_names and raise a different warning/error in checks than the case of a fixed value.

phoebe/atmospheres/passbands.py
```diff
@@ -716,7 +721,7 @@ def load(cls, archive, load_content=True):
                 stored_atms = set([content.split(':')[0] for content in self.content])
 
                 # TODO: replace with < parse('2.5.0') when 2.5.0 is released
-                if parse(self.phoebe_version) != parse('2.4.17.dev+feature-blending'):
+                if parse(self.phoebe_version) != parse('2.4.22.dev+feature-blending'):
```
aprsa: we need to be a little bit careful about this: we are matching a version of phoebe that computed passbands here, and the committeed passbands are using 2.4.21.dev, not (yet) 2.4.22.dev. Thus, this needs to be 2.4.21.dev until we recompute the passbands, or update the version to 2.5.

Raindogjones: Ah, yeah, I'll revert this.

kecnry:
has anything changed that would affect the extinction test? (Should we just loosen the tolerances slightly or is this a cause for concern?)

tests/tests/test_extinction/test_bb_extinction.py
```diff
@@ -51,7 +56,7 @@ def test_bb_extinction():
         extrapolation_method='none'
     ).get_interpolated_values().flatten()
 
-    assert np.allclose(iext, iext_predicted, atol=2e-3, rtol=2e-3)
```
kecnry, line 54: this tolerance change allows tests to pass - is there a difference in the passbands themselves or is it possible trapz changed slightly?

aprsa: The issue is intensity weighting: the test compared energy-weighted intensities with photon-weighted tables. That worked before because intensity weighting was inverted. Now that that's fixed, this test was failing. The simplest fix is to change intensity weighting from 'photon' to 'energy' and the test passes.

aprsa (approved):
I believe we're good to go!

PR #1121, opened 2026-05-05
Moves stock passband logic to dedicated CLI.

aprsa:
* adds a phoebe/cli/ directory for all command-line interface scripts;
* adds admin.py CLI script that can be used to create stock passbands;
* hooks admin.py to phoebe-admin executable in pyproject.toml. To run, do `phoebe-admin create stock-passbands` in the directory above top-level phoebe/;
* adds default transmission files for Johnson V and bolometric passbands;
* updates bolometric and Johnson V passbands;
* renames get_available_atms() to get_supported_atms() and expands its functionality;
* modifies endpoints for the passband server and local passband directory to simplify testing.


Raindogjones:
I'm not sure I understand the point of `admin.py`, shouldn't we default to downloading from the passband server rather than creating them on the fly? Creating them on the fly requires access to the ck2004 and phoenix tables, and we don't want to package these with phoebe as they are even heavier than the passbands themselves, right?

aprsa:
Correct; phoebe-admin is only for devels to create stock passbands when necessary. This is a substitute for the dead code that used to live in passbands.py. This should be opaque to users (and most devels, except you and me, in all likelihood).

Raindogjones (commented):
Looks good to me then!

phoebe/atmospheres/passbands.py
```diff
@@ -42,7 +42,7 @@
 
 # define the URL to query for online passbands.  See tables.phoebe-project.org
 # repo for the source-code of the server
-_url_tables_server = os.getenv('PHOEBE_TABLES_SERVER', 'https://tables.phoebe-project.org')
+_url_tables_server = os.getenv('PHOEBE_TABLES_SERVER', 'https://tables25.phoebe-project.org')
```
kecnry: we will need to revert this once we do the redirection work on the server-side

kecnry (approved):
anything else from our plan re passband server that needs to be done now before merge?  Or are those all waiting on the changes to the server first?

kecnry:
@aprsa - can you handle the conflicts between this and release-2.5 please?  (My suggestion would be to reset to release-2.5 and then cherrypick the one commit over there, seems there are just a few conflicts)

phoebe/atmospheres/passbands.py
kecnry: this should now fix itself on the rebase

aprsa:
Done -- I cherry-picked the commit, fixed the conflicts, and force-pushed back to the pb_cli branch. The tests are now passing, conflicts are resolved, and I believe we're good to merge.

kecnry (approved):

PR #1122, opened 2026-05-07
Returns interval weighted LD coefficients

Raindogjones:
I *think* this brings back interval weighted limb darkening coefficients, which are needed for WD atmospheres.

### DIFF
--- a/phoebe/atmospheres/passbands.py
+++ b/phoebe/atmospheres/passbands.py
@@ -935,7 +935,7 @@ class Passband:
         else:
             raise NotImplementedError(f'ld_func={ld_func} is not supported.')
 
-    def compute_intensities(self, atm, include_mus=True, include_ld=True, include_extinction=False, rvs=None, ebvs=None, add_history_entry=True, verbose=True):
+    def compute_intensities(self, atm, include_mus=True, include_ld=True,  ld_weighting='uniform', include_extinction=False, rvs=None, ebvs=None, add_history_entry=True, verbose=True):
         """
         Computes direction-dependent passband intensities using the passed `atm`
         model atmospheres.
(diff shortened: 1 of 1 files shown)


phoebe/atmospheres/passbands.py
```diff
@@ -1127,18 +1129,27 @@ def compute_intensities(self, atm, include_mus=True, include_ld=True, include_ex
             # define the residuals function for the least squares optimization:
             def ld_resids(x, *args, **kwargs):
                 xdata = kwargs.get('xdata')
+                
+                if ld_weighting=='uniform':
+                    sigma=np.ones(len(xdata))
+                elif ld_weighting=='interval':
+                    delta = np.concatenate( (np.array((xdata[1]-xdata[0],)), xdata[1:]-xdata[:-1]) )
```
aprsa: this might look a tad nicer like this:
```
diffs = np.diff(xdata)
delta = np.r_[diffs[0], diffs]
```
but it accomplishes the exact same thing.

aprsa (approved):
Looks good to me!

aprsa:
It just dawned on me that we also need to support passing LD weighting to compute_intensities(), either via kwargs (currently unsupported) or via a named argument. Otherwise LDs would need to be computed post-hoc always.

Raindogjones:
> It just dawned on me that we also need to support passing LD weighting to compute_intensities(), either via kwargs (currently unsupported) or via a named argument. Otherwise LDs would need to be computed post-hoc always.

I'm a bit confused. Isn't that what I've implemented? When I recalculated the passband for the paper plots I just passed `ld_weighting='interval'` as a kwarg to compute_intensities()

aprsa:
> > It just dawned on me that we also need to support passing LD weighting to compute_intensities(), either via kwargs (currently unsupported) or via a named argument. Otherwise LDs would need to be computed post-hoc always.
> 
> I'm a bit confused. Isn't that what I've implemented? When I recalculated the passband for the paper plots I just passed `ld_weighting='interval'` as a kwarg to compute_intensities()

You're right, it's already done in this PR -- I was looking at my own version as I'm gearing to recompute the LDs and the DA tables.

Raindogjones:
Clearly sleep deprived! If there is anything I can do to lighten the load, please let me know!

phoebe/atmospheres/passbands.py
Raindogjones: Changed, thanks!

PR #1141, opened 2026-07-01
update b.references for 2.5 release and accompanying paper

kecnry:
currently pointing to the not-yet-live landing page on phoebe docs.  Once the publication is actually live, we can replace with a link to ADS (in whatever is the next bugfix after that).

**NOTE**: I also updated the Kochoska in prep citations since that paper is no longer planned, but wondering if we should either remove those or point them to this paper instead?  Or we can flesh out the page that it points to on the website with some information (but ultimately there is no DOI able to be cited, so we should probably give some guidance on what if anything to cite, or just remove those entries entirely to fallback on the original phoebe paper(s)).


aprsa:
I would remove the entries entirely -- for fitting/best practices, I'd point to Conroy et al.

PR #1143, opened 2026-07-02
blending_factors/extrapolation_dists map to nans after vert>face

kecnry:
This PR changes when zeros are mapped to nans for the mesh-exposed values for blending_factors and extrapolation_dists to be _after_ averaging the vertex values into a face-value.  Previously, a zero for any vertex (no extrapolation) would map to nan (no extrapolation) for the entire face because we were mapping zeros to nans before getting the element central value.

Note that estimating which triangles should have extrapolation/blending based on any input parameter to atmospheres (temperature, logg, etc) may not yield the same number of triangles as it is possible that the central value of a triangle is within bounds, even though one of the vertices was not.

In other words:

```
extrapolation method = linear
   N teff < 3500: 52
   N extrapolated: 67
   N blended: 67
```

does make sense, as the remaining 15 triangles had an "average" temperature in-bounds, but at least one vertex had a temperature (or potentially some other atmosphere parameter) out-of-bounds.

### DIFF
--- a/phoebe/backend/backends.py
+++ b/phoebe/backend/backends.py
@@ -1442,6 +1442,11 @@ class PhoebeBackend(BaseBackendByTime):
 
                             value = body.mesh[key].centers
 
+                            if indep in ['blending_factors', 'extrapolation_dists']:
+                                # Apply on-grid masking at face-level so partially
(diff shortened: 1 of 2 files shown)


aprsa (approved):
looks good to me!

PR #1146, opened 2026-07-08
extrapolation_frac_allowed parameter

kecnry:
This parameter limits the number of allowed elements that use extrapolation and raises an error, within `run_compute` (since it requires the mesh to be populated), if that fractional value is exceeded.  The goal here is to allow some extrapolation out-of-the-box but require opt-in for full off-grid fallback.

If accepted, this probably would require some updates to the blending tutorial (and might motivate a change in the default behavior of extrapolation and/or blending method parameters).


phoebe/backend/backends.py
```diff
@@ -992,6 +992,17 @@ def _worker_setup(self, b, compute, times, infolists, **kwargs):
         hier = b.get_hierarchy()
         starrefs  = hier.get_stars()
         meshablerefs = hier.get_meshables()
+        extrapolation_frac_allowed_override = kwargs.get('extrapolation_frac_allowed', None)
+        extrapolation_frac_allowed = {
+            component: computeparams.get_value(
+                qualifier='extrapolation_frac_allowed',
+                component=component,
+                extrapolation_frac_allowed=extrapolation_frac_allowed_override,
+                default=1.0,
```
Raindogjones: Shouldn't this default be 0.1 rather than 1.0?

Raindogjones (commented):
Looks good to me but I haven't tested locally yet.

phoebe/backend/backends.py
kecnry: yes (although shouldn't affect functionality unless the parameter was missing)

PR #1161, opened 2026-08-11
fixes array dimensionality quirk in legacy passband generation

aprsa:
When exporting to pre-2.5 passband format, blackbody grid had to be squeezed before fitting a spline; fixed.

tests/local sneaked into .gitignore to reduce chatter.


kecnry (commented):
can you update with the branch and add a changelog entry (or let me know what it should say and I can handle adding it)?

And is there any way to catch this in a regression test or not worth it?

PR #1170, opened 2026-08-13
fixes a segfault when distortion_method='none' for both components

aprsa:
The issue was in passing non-meshed arrays to libphoebe for reflection, which caused a segfault. This is now fixed on two levels: run_checks() now verifies whether bodies have associated fluxes, and python-side reflection now checks whether distortion method warrants reflection computation. While at it, Claude was kind enough to write a test suite for this so that we are regression-proof.


kecnry (approved):
I was going to try to see if the checks could be consolidated... but let's at least get this in for now since it fixes the bug

